Issue #24, opened 2019-04-04
Permission system first revision


monkeydom:
With the users from #9 implement a permission system working for the API and interface for Podcasts and Episodes

PR #39, opened 2019-04-17
The write side for #24: permissions on all entities and users, and possibility to set and read them

monkeydom:
This is a partial pull request for now, maybe I'll finish the read side today too, but it seems to me that it would be less integration work if we pull that part in earlier if I still need more time.


guides/Data Model and Structure.md
```diff
@@ -0,0 +1,41 @@
+# Data Model and Structure
```
monkeydom: Just an in progress file that should end up in guides eventually.

priv/repo/migrations/20190409070736_add_perm_tables.exs
```diff
@@ -0,0 +1,41 @@
...
+      add :user_id, references("auth_users", on_delete: :delete_all), primary_key: true
+      add :episode_id, references("episodes", on_delete: :delete_all), primary_key: true
+      add :permission, :string, size: 16, null: false
+
+      timestamps()
+    end
+
+    flush()
+
+    Radiator.Repo.insert(Radiator.Auth.User.reserved_user(:public))
```
monkeydom: My initial idea was to use the rights of the reserved_user(:public) to define the public published queries to make sure only stuff is public that is. Don't know if that is a good idea anymore, but it is still around for now until I have the read side of the permissions going.

lib/radiator/auth/user.ex
```diff
@@ -47,6 +17,21 @@ defmodule Radiator.Auth.User do
     timestamps()
   end
 
+  @doc """
+  this is the upper bound of reserved `user_id`s. All regular users need to have an id above this.
+  """
+  def max_reserved_user_id, do: 10
+
+  def reserved_user(:public) do
+    %User{
```
monkeydom: See migration for rationale.

lib/radiator/directory.ex
```diff
@@ -1,14 +1,13 @@
...
   """
 
   import Ecto.Query, warn: false
 
-  alias Radiator.Directory.Episode
-  alias Radiator.Directory.Podcast
-  alias Radiator.Directory.Network
+  alias __MODULE__
   alias Radiator.Repo
+  alias Directory.{Network, Episode, Podcast}
```
monkeydom: General style question here: should aliases already include previous aliases? I tend to like it and have some changes here. E.g.

```elixir
alias Radiator.Directory
alias Directory.Episode
```
vs
```elixir
alias Radiator.Directory
alias Radiator.Directory.Episode
```

priv/repo/migrations/20190409070736_add_perm_tables.exs
eteubert: Interesting. A downside I see is that one _could_ give this user more rights and suddenly everyone has more rights -- opposed to hardcoding public access.

lib/radiator/directory.ex
```diff
@@ -263,68 +123,5 @@ defmodule Radiator.Directory do
...
-  ## Examples
-
-      iex> change_network(network)
-      %Ecto.Changeset{source: %Network{}}
-
-  """
-  def change_network(%Network{} = network) do
-    Network.changeset(network, %{})
-  end
+  ## Todo: Needs to move to `Editor` context
```
eteubert: is this code comment outdated?

lib/radiator/directory/editor.ex
```diff
@@ -9,28 +9,26 @@ defmodule Radiator.Directory.Editor do
...
+
+  def get_permission(user = %Auth.User{}, subject = %Podcast{}),
+    do: get_permission_p(user, subject)
+
+  def get_permission(user = %Auth.User{}, subject = %Episode{}),
+    do: get_permission_p(user, subject)
 
-  def get_permission(user = %Auth.User{}, subject = %Network{}) do
-    case Repo.get_by(NetworkPermission, user_id: user.id, network_id: subject.id) do
+  defp get_permission_p(user, subject) do
```
eteubert: The `_p` suffix is for "private"? The convention is to prefix the name with "do_" if you need to avoid name clash like here: `defp do_get_permission()`.

https://github.com/christopheradams/elixir_style_guide#private-functions-with-same-name-as-public

eteubert:
On first glance the polymorphic approach looks better 👍

priv/repo/migrations/20190409070736_add_perm_tables.exs
monkeydom: Well, in my head it would also give another context and frontend, but use mainly the same stuff underneath. Maybe we can get best of both worlds here without accidentally giving the public more access. But then again it might be more annoying to exclude those users in all other interfaces. Let's see.

lib/radiator/directory.ex
monkeydom: Okay, then I'll stop this for now. My dislike for redundancy produced this.

lib/radiator/directory.ex
monkeydom: Yes.

lib/radiator/directory/editor.ex
monkeydom: Okay, I can go with that convention.

Issue #20, opened 2019-03-26
Podcast & Episode Slugs


eteubert:
When an episode or podcast is created, a "slug" must be generated and saved in the database. Slugs must be simple and URL-safe, so only a-z0-9 and dashes.

```elixir
slug "Is this the end of the wōrld?"
"is-this-the-end-of-the-world"
```

Slugs must _not_ be changed on update as this would lead to changing URLs.

Ideally slugs are not written on create but the first _publishing_ of an entity to allow for preliminary draft titles.

PR #43, opened 2019-04-18
WIP: Podcast & Episode Slugs

optikfluffel:
I gave #20 a first shot and want to ask some feedback before continuing.

Also regarding Episodes: it seems there isn't any publishing mechanism (yet?), should I wait for that, add it myself, or just generate the slug when creating the Episode?


priv/repo/migrations/20190418150514_add_slug_to_podcast_table.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Radiator.Repo.Migrations.AddSlugToPodcastTable do
+  use Ecto.Migration
+
+  def change do
+    alter table(:podcasts) do
+      add :slug, :string
```
eteubert: Feel free to add this to the original podcast creation migration instead of its own migration. We recently decided to adjust existing migrations for now instead of adding new ones.

lib/radiator_web/resolvers/directory.ex
```diff
@@ -69,6 +69,22 @@ defmodule RadiatorWeb.Resolvers.Directory do
       nil ->
         {:error, "Podcast ID #{id} not found"}
 
+      %Podcast{slug: nil} = podcast ->
```
eteubert: My main concern with your current implementation is here. You implemented the slugging in the GraphQL resolver, so it relies on GraphQL being used -- which is an abstraction layer too high. It should apply when the main elixir API, `Directory.update_podcast/2`, is called. I suggest doing it inside the podcast changeset. 

Maybe https://github.com/sobolevn/ecto_autoslug_field is worth a closer look. I don't know if it's flexible enough for our use case but at least the slugging takes place in the right spot.

lib/radiator_web/resolvers/directory.ex
```diff
@@ -69,6 +69,22 @@ defmodule RadiatorWeb.Resolvers.Directory do
...
+          published_at: DateTime.utc_now(),
+          slug: slug
+        }
+
+        case Directory.update_podcast(podcast, attrs) do
+          {:ok, podcast} ->
+            {:ok, podcast}
+
+          {:error, %Ecto.Changeset{errors: [slug: {"has already been taken", _}]}} ->
+            {:error, "Slug #{slug} has already been taken"}
```
eteubert: I think we should try to generate unique slugs and not reject them, meaning: saving of a podcast (or any other sluggable entity) should never fail due to a unique slug constraint.

I suggest copying what WordPress does: appending an incrementing number until a unique slug is found. `foo-bar` -> `foo-bar-1` -> `foo-bar-2` etc.

lib/radiator_web/resolvers/directory.ex
optikfluffel: Ok, putting it into `Radiator.Directory` was my initial thought. But when I was looking for `publish_podcast/3` and `depublish_podcast/3` I found them in `RadiatorWeb.Resolvers.Directory`. Should I move these two functions as well, so they are not only available in/for GraphQL?

optikfluffel:
@eteubert two more questions:

* Can I just add a bunch of commits to resolve everything and you squash when merging, or should I try to keep the history "clean" and amend to the last commit?
* Is there a reason Travis isn't running for Pull Requests?

eteubert:
@optikfluffel 

Commit to your liking, we squash-merge in the end.

I think I fixed travis. Must have been some mixup when renaming the repo.

lib/radiator_web/resolvers/directory.ex
eteubert: Right, that would be good!

eteubert:
Nearly there :) One last quibble: `publish_podcast` and `depublish_podcast` should take a `%Podcast{}` argument, not an id, similar to the existing `update_podcast`, `delete_podcast` etc.

optikfluffel:
Anything else? ^^

If not, I'd like to tackle the Episodes somewhen during the weekend. What about `publish_episode/1` & `depublish_episode/1`? Should I just add those as well?

eteubert:
Looks good :) 

> What about publish_episode/1 & depublish_episode/1? Should I just add those as well?

👍🏻

eteubert:
⚠️  Latest commit (232a182dab2b8382cbd976a0d0a4143ded8bc990) moved `Directory` API around a bit. `Radiator.Directory` should now only be used for accessing public data, everything requiring an authorized user lives in `Radiator.Directory.Editor`.

lib/radiator/directory.ex
```diff
@@ -54,6 +54,16 @@ defmodule Radiator.Directory do
   def get_podcast!(id), do: Repo.get!(Podcast, id)
   def get_podcast(id), do: Repo.get(Podcast, id)
 
+  @doc """
+  Gets a single podcast by it's slug.
```
monkeydom: typo: its

optikfluffel:
Do the Episodes also need the `field :is_published, :boolean` in the GraphQL API?

eteubert:
Yes that would make sense.

I just realize that networks should also have a slug. However I don't think networks have a "published" state, they simply exist, so the slug must be generated on create. Do you mind adding this? Alternatively I can create a new issue for the new requirement.

optikfluffel:
I gave it a try and guess it's time for another round of feedback✨.

PR #49, opened 2019-04-24
feat(api,auth,perm): add api to generate access token with credentials, use these credentials in an api call

### LINKED ISSUE #45: Add API for potential frontends to get a Access Token with credentials
To unblock frontend development

### LINKED ISSUE #46: Take Access Token into account for API requests
Depends on #45

monkeydom:
* the `authenticatedSession` mutation now returns a `session` object with a token that needs to be put into the `Authorization:` header with a `Bearer ` prefix to have authenticated api calls.
* put Authentication in place for the `createNetwork` mutation. Default TTL is 45 minutes.
* add a `prolongSession` mutation to get a new token with full TTL to not need credentials again in a potential frontend app that is in use. 
* update `network_tests` to use this. 
* add `users_tests` for session related tests
* move middleware to `Radiator.Schema.Middleware`, rename `ChangesetMiddleware` to `Middleware.TranslateChangeset`

Fixes #45 
Fixes #46

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,20 @@ Entrypoint: `/api/graphql`
 
 Open http://localhost:4000/api/graphiql for schema and documentation exploration.
 
+For calls that need authentication, make sure to put the token gotten from a 
+
+```GraphQL
+mutation { 
+	authenticatedSession(
+		username_or_email: "admin", 
+		         password: "password" ) { 
+		token 
+	} 
+}
+```
+
+request into the `Authorization: Bearer <token>` header.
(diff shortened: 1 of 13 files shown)


eteubert:
Is the ttl visible to clients? Maybe excplicitly include the expiry timestamp in the graphql session object?

monkeydom:
Good question. The expiry is in the data payload under the default 'exp' claim. However, I could easily also expose this if needed in the api, but I currently would wait for a client to

PR #55, opened 2019-05-05
Log emails to console in dev profile

djschilling:
This pr introduces a new Bamboo.Adapter which logs the emails to the console.
It also activates it for the dev profile.

The idea behind it is to be able to see the sent emails immediately in the console without having to configure an smtp server for development.

I'm absolutely new to Elixir and this tech stack. So if there are things completely wrong, please tell me.

I was not sure if i should use `IO.puts` to print to the console. But i didn't find other logging things in the project so i went with it.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -74,3 +74,6 @@ config :radiator, Radiator.Repo,
   database: "radiator_dev",
   hostname: "localhost",
   pool_size: 10
+
+config :radiator, Radiator.Mailer,
+  adapter: Radiator.Email.Console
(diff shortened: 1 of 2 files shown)


eteubert:
Very useful, thank you.

I'd say using the [Logger](https://hexdocs.pm/logger/Logger.html) would be good, probably [Logger.info](https://hexdocs.pm/logger/Logger.html#info/2).

Please use [automatic code formatting](https://elixir-lang.org/getting-started/mix-otp/introduction-to-mix.html#automatic-code-formatting). It is recommended to auto-run this on save.

djschilling:
- [x] Logger
- [x] Format

eteubert:
Thanks :)

PR #56, opened 2019-05-07
feat(auth,user): make Register.get_user_by_credentials/2 case insensitive

### LINKED ISSUE #40: Login by name/email should be case insensitive
Can be done by adding a lower case index on these columns and querying against that column.

```elixir
create index(:users, ["(lower(email))"], unique: true)
```

via https://minhajuddin.com/2019/04/14/how-to-store-username-or-email-with-case-insensitive-search-using-ecto/

optikfluffel:
Closes #40 when accepted and reviewed.


eteubert:
Thanks :)

PR #58, opened 2019-05-07
Use the default constraints for signup validation (email, name)

### LINKED ISSUE #57: Ecto.ConstraintError on signup
When providing the same name as someone else did before in signup an `Ecto.ConstraintError` shows up. The same problem exists when using the same email as someone else used before.

Screenshot 2019-05-07 at 22 08 40

A validation message should show up here instead of the server error.

djschilling:
The validation was wrongly changed when changing the unique constraint
to be case insensitive here: https://github.com/podlove/radiator/pull/56/commits/23cbfe57e3c9e5cfbac9582f73c0bc3c80e342da

Closes #57

### DIFF
--- a/lib/radiator/auth/user.ex
+++ b/lib/radiator/auth/user.ex
@@ -40,8 +40,8 @@ defmodule Radiator.Auth.User do
     user
     |> cast(attrs, [:name, :email, :display_name, :password, :password_hash, :status])
     |> cast_attachments(attrs, [:avatar])
-    |> unique_constraint(:name, name: :auth_users__lower_name_index)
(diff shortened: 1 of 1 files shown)


optikfluffel:
@djschilling Hey 👋, I'm sorry if I did this wrong, but when playing around with it I didn't get any validation errors when trying to add a second user with a differently cased, but otherwise equal, name/email without those changes. I probably should have added another test for that to be sure 🤔.

djschilling:
Okay, i just tested it end after a `mix ecto.reset` it is working without my fix. so I'm closing this.
@eteubert thanks for clarification.

PR #67, opened 2019-05-15
feat(api,graphql): add episodesCount to podcast GraphQL queries

### LINKED ISSUE #62: GraphQL: add field for total number of episodes on a podcast
A user needs to know the number of episodes in a podcast.

Not sure if there is a naming convention for this in GraphQL, otherwise I suggest `episodes_count`.

optikfluffel:
I'm not too happy with making an extra database query to get the `episode_count`. Open to suggestions for a better idea. 😋

---

closes #62, when merged

### DIFF
--- a/lib/radiator/directory.ex
+++ b/lib/radiator/directory.ex
@@ -71,24 +71,14 @@ defmodule Radiator.Directory do
   @doc """
   Gets the number of episodes of the podcast with the given id.
 
-  Raises `Ecto.NoResultsError` if the Podcast does not exist.
-
   ## Examples
 
       iex> get_episodes_count_for_podcast!(123)
       3
-
-      iex> get_podcast!(456)
-      ** (Ecto.NoResultsError)
-
   """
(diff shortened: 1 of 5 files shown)


lib/radiator/directory.ex
```diff
@@ -68,6 +68,29 @@ defmodule Radiator.Directory do
...
+  ## Examples
+
+      iex> get_episodes_count_for_podcast!(123)
+      3
+
+      iex> get_podcast!(456)
+      ** (Ecto.NoResultsError)
+
+  """
+  def get_episodes_count_for_podcast!(id) do
```
eteubert: I think a plain Ecto query is more readable here:

```elixir
Repo.aggregate((from e in Episode, where: e.podcast_id == ^id), :count, :id)
```

or

```elixir
(from e in Episode, where: e.podcast_id == ^id)
|> Repo.aggregate(:count, :id)
```

PR #85, opened 2019-05-30
feat(import): improve import to also take enclosures, show and episode titles

### LINKED ISSUE #74: Improve import of existing podcasts
E.g.:
* show progress while importing
* download and upload all media
* use the chapter parsing to extract chapter information from mp3 id3 if there

monkeydom:
add(phx-interface): delete podcast, show chapter links and times, show episode images, show podcast images
fix(arc): fix setting of images on podcast creation (was missing the id during the changset building when new, same is still true for episodes)
fix(arc): add sideloading for enclosures (doing so inline did cause a transaction timeout)
refactor(import): moved import into import.ex
feature(import): import first 10 enclosures now, also take images again
fix(phx-interface): ensure audio is shown again if present

needs ecto.reset so deletions are working
fixes #74


lib/radiator/directory/importer.ex
```diff
@@ -0,0 +1,125 @@
...
+
+  def import_enclosures(
+        user = %Auth.User{},
+        podcast = %Podcast{},
+        feed = %Metalove.PodcastFeed{},
+        limit \\ 10
+      ) do
+    Metalove.PodcastFeed.trigger_episode_metadata_scrape(feed)
+    Logger.info("Import: Scraping metadata for #{feed.feed_url}")
+    feed = Metalove.PodcastFeed.get_by_feed_url_await_all_metdata(feed.feed_url, 1_000 * 15 * 60)
```
eteubert: Gibt für solche Zeitangaben einen nützlichen Helper in Erlang

```
iex(1)> :timer.minutes(1)
60000
iex(2)> :timer.minutes(15)
900000
```

PR #108, opened 2019-06-10
Merge public+admin GraphQL schemas into one

### LINKED ISSUE #107: Solve GraphQL public/admin schema separation
Currently there are two schemas, `RadiatorWeb.GraphQL.Public.Schema` and `RadiatorWeb.GraphQL.Admin.Schema`. This makes discovery hard.

There is an [Absinthe ticket for schema stitching](https://github.com/absinthe-graphql/absinthe/issues/387) but that is unresolved.

Maybe we can do something as simple as prefixing all public queries with `public_` and call it a day.

eteubert:
- merged both schemas into one
- prefixed all public objects with "public" to work around name conflicts and to make it obvious which is which
- added `RadiatorWeb.GraphQL.Common` module for objects that are (currently) identical in public and admin

fixes #107


eteubert:
Removed the prefix in field/attribute names, feels good to me. The only place where I left it is `publishedEpisodesCount` because that just describes what's in the tin :)

monkeydom:
``` query { networks {
 id 
  podcasts {
    id
    title
    subtitle
  }
}
} ``` leads to an exception, but should just lead to unauthorized, same with network(id: 1). Everything works with a correct bearer token

eteubert:
👍  Middleware was missing

monkeydom:
Looks good to me now.

PR #111, opened 2019-06-12
feature(api): add `userSignup` to graphQL api

### LINKED ISSUE #75: Add GraphQL mutation for user registration
We have a mutation for login (`authenticatedSession`) but the initial user registration is missing.

monkeydom:
feat(api): add expirydate to session result for `authenticatedSession` and `prolongSession`
fix(api): requiring active users now for admin api

fixes #75

### DIFF
--- a/lib/radiator_web/graphql/admin/resolvers/user.ex
+++ b/lib/radiator_web/graphql/admin/resolvers/user.ex
@@ -1,0 +1,26 @@
+defmodule RadiatorWeb.GraphQL.Admin.Resolvers.User do
+  alias RadiatorWeb.GraphQL.Helpers.UserHelpers
+  alias Radiator.Auth
+
+  def resend_verification_email(_parent, _params, %{
+        context: context
+      }) do
+    case Map.get(context, :authenticated_user) do
+      %Auth.User{status: :unverified} = user ->
+        case UserHelpers.resend_verification_email_for_user(user, context) do
+          :sent ->
+            {:ok, true}
+
+          # Consideration: Maybe make it an error instead?
+          _ ->
+            {:ok, false}
+        end
+
+      _ ->
+        {
+          :error,
+          "Authenticiation: Needs authenticated unverified user."
+        }
+    end
+  end
+end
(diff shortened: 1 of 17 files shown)


lib/radiator_web/graphql/public/resolvers/session.ex
```diff
@@ -1,21 +1,63 @@
...
-         }}
+      {:error, _changeset} ->
+        {:error, "Failed to create #{username} <#{email}>"}
     end
   end
+
+  @doc """
+  Helper method to be shared
+  """
+  def new_session_for_valid_user(user) do
```
eteubert: I feel like this belongs in `Radiator`, not `RadiatorWeb`.

lib/radiator_web/graphql/helpers/user_helpers.ex
```diff
@@ -0,0 +1,39 @@
...
+            resolution_context.context_conn,
+            user
+          )
+        )
+        |> Radiator.Mailer.deliver_later()
+
+        :sent
+
+      _ ->
+        :not_sent
```
eteubert: Maybe be more specific here like `:not_sent_already_verified`. Out of context `:not_sent` reads like a technical reason, not a logical one.

monkeydom: well, I left it  vague for now as it also covers the `:suspended` case I have a state so it is quite vague. We just have to revisit this after we actually reuse the different status options other than `:unverified`

lib/radiator_web/graphql/admin/resolvers/user.ex
```diff
@@ -0,0 +1,26 @@
...
+        context: context
+      }) do
+    case Map.get(context, :authenticated_user) do
+      %Auth.User{status: :unverified} = user ->
+        case UserHelpers.resend_verification_email_for_user(user, context) do
+          :sent ->
+            {:ok, true}
+
+          # Consideration: Maybe make it an error instead?
+          _ ->
```
eteubert: This branch can never be called because you already match on `%Auth.User{status: :unverified}` at line 9

PR #117, opened 2019-06-13
feat: add support for Podlove Web Player

### LINKED ISSUE #80: Podlove Web Player integration
- [x] add web player to project
- [x] add configuration endpoint for episode player configs

eteubert:
- add config URLs for audios and episodes
- display web player on episode page instead of HTML5 audio element

fixes #80

### DIFF
--- a/lib/radiator_web/controllers/player_controller.ex
+++ b/lib/radiator_web/controllers/player_controller.ex
@@ -1,0 +1,48 @@
+defmodule RadiatorWeb.PlayerController do
+  use RadiatorWeb, :controller
+
+  alias Radiator.Directory
+  alias Radiator.Directory.Audio
+  alias Radiator.Media.AudioFile
+  alias Radiator.AudioMeta.Chapter
+
+  def show(conn, %{"audio_id" => id}) do
+    audio = Directory.get_audio(id)
+
+    json(conn, config(conn, audio))
+  end
+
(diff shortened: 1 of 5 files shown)


lib/radiator_web/router.ex
```diff
@@ -38,6 +38,9 @@ defmodule RadiatorWeb.Router do
 
     get "/feed/:podcast_id", FeedController, :show
 
+    get "/audio/:audio_id/player.json", PlayerController, :audio_config
+    get "/episode/:episode_id/player.json", PlayerController, :episode_config
```
monkeydom: imho those routes should not exist this way. we need a public route infrastructure, and these need to be under whatever the podcast/episode route is. But probably okay for now, maybe add a comment? I would also like us very much to add some indirection to the id for public urls. E.g. Either have a radical uuid to internal id mapping, or something more semantic and nice. But exposing the internal database id's for the public pages I'm not happy with.

lib/radiator_web/controllers/player_controller.ex
```diff
@@ -0,0 +1,80 @@
+defmodule RadiatorWeb.PlayerController do
```
monkeydom: just as a work item: we should sit together and unify the keys/values for the config and our graphql schema to make this as frictionless and meaningful as possible.

Issue #27, opened 2019-04-04
Add user settings screen


monkeydom:
* change password
* add avatar
* change display name
* change email (more elaborate, needs confirmation of new email by sending a mail to the old account before it sticks)

djschilling:
Should the frontend be implemented with Vue.js as mentioned in the README or should serverside rendering be used like in https://github.com/podlove/radiator/blob/master/lib/radiator_web/templates/login/signup_form.html.eex

PR #118, opened 2019-06-14
Password Change

djschilling:
Implement the first part of #27.

### DIFF
--- a/lib/radiator/auth/user.ex
+++ b/lib/radiator/auth/user.ex
@@ -52,7 +52,7 @@ defmodule Radiator.Auth.User do
     |> validate_required([:password_hash])
   end
 
-  def change_password_changelog(%User{} = user, attrs) do
+  def change_password_changeset(%User{} = user, attrs) do
     user
     |> cast(attrs, [:password, :password_confirmation])
     |> validate_required([:password, :password_confirmation])
--- a/lib/radiator/auth/user.ex
+++ b/lib/radiator/auth/user.ex
@@ -14,6 +14,7 @@ defmodule Radiator.Auth.User do
     field :avatar, Radiator.Media.UserAvatar.Type
     field :password_hash, :binary
     field :password, :string, virtual: true
+    field :password_confirmation, :string, virtual: true
     field :status, Radiator.Auth.Ecto.UserStatusType, default: :unverified
     # unverified, active, suspended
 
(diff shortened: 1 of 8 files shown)


lib/radiator_web/controllers/admin/usersettings_controller.ex
```diff
@@ -0,0 +1,45 @@
...
+  alias Radiator.Auth
+
+  action_fallback RadiatorWeb.Api.FallbackController
+
+  def index(conn, _params) do
+    render(conn, "usersettings.html", changeset: Auth.Register.change_user(%Auth.User{}))
+  end
+
+  def create(conn, params) do
+    user_map = params["user"]
```
eteubert: `user_params` is the more common variable name choice here. Variable type is not usually included in the name.

lib/radiator_web/controllers/admin/usersettings_controller.ex
```diff
@@ -0,0 +1,45 @@
...
+
+    cond do
```
eteubert: If all conditions of cond are true and false, use if/then/else.

lib/radiator_web/controllers/admin/usersettings_controller.ex
```diff
@@ -0,0 +1,45 @@
...
+
+  alias Radiator.Auth
+
+  action_fallback RadiatorWeb.Api.FallbackController
+
+  def index(conn, _params) do
+    render(conn, "usersettings.html", changeset: Auth.Register.change_user(%Auth.User{}))
+  end
+
+  def create(conn, params) do
```
eteubert: Maybe `update`? You're not creating anything.

djschilling: I saw that the `resources` function was used in the admin routes. And there i think the create method which stands for `POST /admin/usersettings` is the best match.

Should i switch the routes to not use the `resources` function, and define the http methods directly?

eteubert: You probably need more flexibility here than the `resources` macro provides. It's ok to define custom routes and ideal http methods.

djschilling:
All mentioned things are fixed

eteubert:
One way to simplify the controller update logic is `with`, which lets you focus on the happy path first and deal with errors later. Something like this:

```elixir
    with {:ok, _} <- User.check_password(user, user_params["password_current"]),
         {:ok, _user} <- Auth.Register.update_user(user, user_params) do
      conn
      |> put_flash(:info, "Successfully updated password.")
      |> redirect(to: Routes.admin_network_path(conn, :index))
    else
      # handle error cases here
    end
```

Writing a custom Ecto changeset function may also help increase readability.

djschilling:
That looks nice.

The implementation now uses a seperate changeset and the `with` keyword.

lib/radiator/auth/user.ex
```diff
@@ -45,11 +46,34 @@ defmodule Radiator.Auth.User do
     |> validate_format(:name, ~r/^[^\s ]+$/)
     |> validate_format(:email, ~r/^\S+@\S+$/)
     |> validate_length(:name, min: 2, max: 99)
+    |> validate_length(:password, min: 2)
     |> validate_required([:email, :name])
     |> encrypt_password
     |> validate_required([:password_hash])
   end
 
+  def change_password_changelog(%User{} = user, attrs) do
```
eteubert: typo: changeset, not changelog

eteubert:
Aside from the typo, looks good now 👍

djschilling:
Tipo fixed.

PR #119, opened 2019-06-15
Issue/77 remove bangs

eteubert:
Implemented as discussed but wondering now if that is such a good idea. Phoenix Controllers are designed in a way that raised errors are automatically convertet to appropriate HTTP responses, thereby simplifying controller logic. If we remove the bang getters, we need to implement these cases ourselves, in every controller method.

See https://github.com/podlove/radiator/blob/master/lib/radiator_web/controllers/api/fallback_controller.ex

### DIFF
--- a/lib/radiator/directory.ex
+++ b/lib/radiator/directory.ex
@@ -126,27 +126,6 @@ defmodule Radiator.Directory do
     end)
   end
 
-  @doc """
-  Gets a single published episode.
-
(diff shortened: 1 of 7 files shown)


monkeydom:
Yeah. Did it on my branch too and the read up. I ended up doing a fallback controller which makes also sense for us to bubble up the not authorized in a similar way.

monkeydom:
Also happening in #124 - took your test asserts

PR #120, opened 2019-06-15
People/Contributions

eteubert:
Followed a mood today and spontaneously gave the whole Contributors data stuff a try. The data relationships are modeled and contributors are included in the feed, both for channel and items.

I called the core entity "people" because that's what they are. They only become contributors through their relation to audios or podcasts. 

Not sure about putting people in the `Radiator.Contribution` context as they could exist without Contribution.

There is a relationship between `Person` and `User`. Maybe we can simplify the User model now by relying on attributes in ther Person. Specifically names and avatar.

                          +------+
                          | User |
                          +---+--+
                              |
                              |
                         +----+---+
         +---------------+ Person +-------------+
         |               +--------+             |
         |                                      |
         |                                      |
    +----+----------------+      +--------------+----+
    | PodcastContribution |      | AudioContribution |
    +----+------------+---+      +----+---------+----+
         |            |   +------+    |         |
         |            +---+ Role +----+         |
    +----+----+           +------+           +--+----+
    | Podcast |                              | Audio |
    +---------+                              +-------+

## TODOs

- ~~[ ] Elixir API for people and contributions manipulation~~
- ~~[ ] GraphQL queries and mutations~~ 
- separate issues for API features: #138 #139 
- [x] move some attributes out of User, into Person; always associate a Person when creating a user and probably preload it by default


eteubert:
@monkeydom Not done yet but I think already worth a second set of eyes. Especially the introduction of Person as complement to User.

monkeydom:
I will have a look at it tomorrow. Generally speaking I think the Authentication based data should stay in the Auth.User (e.g. username, display name and email), but I think one user should be able to be tied to one contributor, and e.g. the avatar should live there. And maybe it is good to preload a Person when creating a user with the initial information as you wrote down here.

lib/radiator/feed/episode_builder.ex
```diff
@@ -60,6 +62,18 @@ defmodule Radiator.Feed.EpisodeBuilder do
...
   end
 
+  defp contributors(%Episode{audio: %Audio{contributors: contributors}}) do
+    contributors
+    |> Enum.filter(fn %Person{display_name: name} ->
+      String.valid?(name) && String.length(name) > 0
+    end)
+    |> Enum.map(fn contributor ->
+      element(:"atom:contributor", [element(:"atom:name", contributor.display_name)])
+    end)
```
monkeydom: Since the `atom:contributor` is a `atom:atomPersonConstruct` it also has an uri, email and possible extensions. I think the Person should have at least one uri to actually supply something here. According to `https://tools.ietf.org/html/rfc4287#section-6.4` simple extensions there are possible, so I would encourage the introduction of a `<atom:role>` child element as well.

lib/radiator_web/controllers/feed_controller.ex
```diff
@@ -7,7 +7,7 @@ defmodule RadiatorWeb.FeedController do
   @items_per_page 50
 
   def show(conn, %{"podcast_id" => id, "page" => page}) do
-    podcast = Directory.get_podcast!(id)
+    podcast = Directory.get_podcast!(id) |> Radiator.Repo.preload([:contributors])
```
monkeydom: We probably want a `Directory.get_all_contributiors(podcast)` as well that is the union of the contributors of that podcast and all audio.

In terms of repo preloading: I think we should move those calls to the context, to make the context self contained and outside users not rely on knowing about the repo.

lib/radiator_web/graphql/admin/resolvers/editor.ex
```diff
@@ -302,6 +307,10 @@ defmodule RadiatorWeb.GraphQL.Admin.Resolvers.Editor do
     {:ok, Media.NetworkImage.url({network.image, network})}
   end
 
+  def get_image_url(%User{person: person}, _, _) do
+    {:ok, Media.PersonAvatar.url({person.avatar, person})}
```
monkeydom: I did add an `image_url` function to the ecto schema and bounced through there always to keep the layering more layered.

lib/radiator_web/graphql/admin/types.ex
```diff
@@ -5,6 +5,29 @@ defmodule RadiatorWeb.GraphQL.Admin.Types do
...
+
+    field :display_name, :string do
+      resolve fn user, _, _ -> {:ok, user.person.display_name} end
+    end
+
+    field :nick_name, :string do
+      resolve fn user, _, _ -> {:ok, user.person.nick_name} end
+    end
+
+    field :gender, :string do
```
monkeydom: Wondering if we should leave out the `gender` part, and maybe do a more generic tagging of people that the clients can choose how to fill? e.g. if they want a male/female/cis/trans/nonbinary/etc. tag. Hard baking gender doesn't seem appropriate anymore, even as a freeform text field.

priv/repo/migrations/0500_add_users_table.exs
```diff
@@ -5,8 +5,6 @@ defmodule Radiator.Repo.Migrations.AddUsersTable do
     create table(:auth_users) do
       add :name, :string
       add :email, :string
-      add :display_name, :string
```
monkeydom: I did think the display_name (although not yet fully utilized) is something worth having in the Auth user table. As I wanted the `name` to be fixed per instance, having a way to specify how it is displayed seems nice here. Delegating that to the Person that is used for metadata does not seem right to me.

E.g. when refering to the logged in user, imho these values here should be used and shown, not something that might be changed by someone else than me. (In that case, the avatar also might need to bubble up, however, I did not want to make binary handling a dependency here)

lib/radiator_web/graphql/admin/resolvers/editor.ex
```diff
@@ -302,6 +307,10 @@ defmodule RadiatorWeb.GraphQL.Admin.Resolvers.Editor do
...
+  end
+
   def get_episodes_count(%Podcast{id: podcast_id}, _, %{context: %{authenticated_user: user}}) do
     Editor.get_episodes_count_for_podcast!(user, podcast_id)
   end
```
monkeydom: Hmm - was looking for some sort of api to add contributors here. This is still out, is it? I think even if you do it in a subcontext `Radiator.Contribution` you should expose it here, so we can add permission checks to the modification. Also wondering where to put both the `User` and `Person` admin api's in general?

lib/radiator/auth/register.ex
```diff
@@ -6,11 +6,28 @@ defmodule Radiator.Auth.Register do
...
-  def get_user_by_name(name), do: name |> User.by_name_query() |> Repo.one()
+  def get_user_by_name(name),
+    do:
+      name
+      |> User.by_name_query()
+      |> Repo.one()
+      |> preload()
+
+  def preload(user) do
+    Repo.preload(user, :person)
```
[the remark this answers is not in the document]
eteubert: Right, but you need to be able to get the person for a user -- what's the alternative?

lib/radiator/contribution/person.ex
```diff
@@ -0,0 +1,54 @@
...
+    field :display_name, :string
+    field :nick_name, :string
+    field :gender, :string
+    field :email, :string
+
+    field :avatar, Media.PersonAvatar.Type
+
+    belongs_to :user, User
+
+    timestamps()
```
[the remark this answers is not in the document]
eteubert: Good questions. Network-global maybe, which means they need a network reference. And can only be edited by network owners (except each user who can update his/her own data).

Interesting thought as well about multiple people representing the same entity. My gut feeling is that we want this, but maybe not necessarily right now, to prioritize more urgent features?

lib/radiator_web/controllers/feed_controller.ex
eteubert: Yes, controllers should never call Repo directly. This is temporary. Still good to call it out as "temporary" solutions have that sneaky habit of staying that way for longer than planned ;)

lib/radiator_web/graphql/admin/resolvers/editor.ex
eteubert: Yes, Elixir and gql API are still on the todo, see initial PR description.

lib/radiator/contribution/podcast_contribution.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Radiator.Contribution.PodcastContribution do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  schema "podcast_contributions" do
+    belongs_to :person, Radiator.Contribution.Person
+    belongs_to :podcast, Radiator.Directory.Podcast
+
```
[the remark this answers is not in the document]
eteubert: Both requests implemented in recent commits

lib/radiator/feed/episode_builder.ex
eteubert: Should we have a distinction between public and private emails then? Because the system needs you to supply an email to work but that doesn't necessarily mean that the user wants this email to be public.

lib/radiator_web/graphql/admin/types.ex
eteubert: Removed gender for now and created a separate issue https://github.com/podlove/radiator/issues/126

lib/radiator/auth/register.ex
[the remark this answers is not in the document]
djschilling: I would suggest that we don't need to create a person when registering.
A user should in my opinion not rely on a person.

A person can optionally reference a user so that the system knows when a person is the one currently logged in.

But i would let a user has all properties that are necessary to register.
I think that would be `email`, `username` (not sure if that is necessary), `password`

lib/radiator/contribution/person.ex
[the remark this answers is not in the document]
djschilling: A person could also be in more than one network so that a person can be a contributor in podcasts from different networks and the person can be identified over multiple networks.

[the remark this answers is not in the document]
eteubert: But then who has permissions to edit that person? Just for this we would need to create an entity above networks so we can manage permissions for it. I feel like this creates bigger problems than it solves.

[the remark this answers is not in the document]
djschilling: Thats a good question.

In the best scenario i would say that contributors can edit their own data and we have some kind of unique id for every contributor that can be used so that contributors can be tracked over multiple networks and radiator installations. But thats even farer off for now.

So probably you are right, that currently the best choice is to connect a person (contributor) to one network.

lib/radiator/feed/episode_builder.ex
monkeydom: Yes we do want a distinction. Another reason to keep the account data in the Auth.User and the public facing description of a person in the Person

PR #137, opened 2019-06-26
REST API Rewrite

eteubert:


guides/rest_api.md
```diff
@@ -0,0 +1,44 @@
...
+
+```
+POST /api/rest/v1/auth
+```
+
+### Parameters
+
+| Name                   | Type     | Description                      |
+| ---------------------- | -------- | -------------------------------- |
+| `audio_file[name]`     | `string` | **Required.** Username or email. |
```
monkeydom: `audio_file` needs to be replaced with the proper thing. Nothing?

guides/rest_api.md
```diff
@@ -0,0 +1,44 @@
+# REST API
+
+## Authentication
+
+To make full use of the API you need to authenticate as an existing user. A successful authentication returns a *token* that you can use in subsequent requests to authenticate yourself as that user. You do this by setting the token as request header.
+
+> curl -H "Authorization: Bearer AUTH-TOKEN" http://localhost:4000/api/rest/v1/
+
+```
+POST /api/rest/v1/auth
```
monkeydom: I don't know how rest apis usually handle this, but i would go for `/api/rest/v1/auth/token` instead and keep the top level clear.

lib/radiator_web/controllers/api/audio_file_controller.ex
```diff
@@ -0,0 +1,35 @@
+defmodule RadiatorWeb.Api.AudioFileController do
+  use RadiatorWeb, :controller
+
+  alias Radiator.Media
+
+  action_fallback(RadiatorWeb.Api.FallbackController)
+
+  def create(conn, %{"audio_file" => audio_file_params}) do
+    case Media.create_audio_file(audio_file_params) do
```
monkeydom: since we use the fallback-controller i would suggest using `with` instead of case and let all the errors just fall through and handle upward.

lib/radiator_web/controllers/api/authentication_controller.ex
```diff
@@ -0,0 +1,34 @@
...
+    user = current_user(conn)
+    token = Auth.Guardian.api_session_token(user)
+
+    conn
+    |> json(AuthHelpers.session_response(user, token))
+  end
+
+  def logout(conn, _params) do
+    conn
+    |> Auth.Guardian.Plug.sign_out()
```
monkeydom: sign out doesn't make much sense currently in the api, as the token just expires and we don't store any state. For the UI it is nice to remove it from the session. However, imho the rest api should not be stateful imho.

lib/radiator_web/helpers/auth_helpers.ex
```diff
@@ -0,0 +1,55 @@
...
+  def session_response(user = %User{}, token) do
+    %{
+      username: user.name,
+      token: token,
+      expires_at: Auth.Guardian.get_expiry_datetime(token)
+    }
+  end
+
+  def current_user(conn) do
+    conn.assigns[:current_user]
```
monkeydom: okay, so you finally did go for current_user? I would prefer we unify to `authenticated_user` or `current_user` back in the login controllers.

lib/radiator_web/controllers/api/authentication_controller.ex
```diff
@@ -0,0 +1,34 @@
...
+      user ->
+        token = Auth.Guardian.api_session_token(user)
+
+        conn
+        |> json(AuthHelpers.session_response(user, token))
+    end
+  end
+
+  def prolong(conn, _params) do
+    user = current_user(conn)
```
monkeydom: don't you need to import the AuthHelpers for this to work here? also see the comment on `current_user` vs. `authenticated_user` in the helpers.

eteubert: Because it's so commonly used I auto-import it to all controllers https://github.com/podlove/radiator/pull/137/files#diff-3a959aa961ca1f891ce2804190e66e81

guides/rest_api.md
eteubert: Can you elaborate? I don't follow.

guides/rest_api.md
eteubert: In general, the URL represents the resource and the HTTP method is the _verb_ acting on the resource.

```
GET    /networks lists networks
GET    /networks/1 shows network with id 1
POST   /networks creates a new network
PUT    /networks/1 updates the network
DELETE /networks/1 deletes the network
```

Authentication is a weird one in REST terms as you don't really create something so it's hard to fit into this schema. Maybe a session. We had a discussion about this a while ago https://github.com/podlove/radiator/pull/29#discussion_r272470397

The most REST-ish way I see is POST to `/sessions` to create a session to log in. Not sure about prolonging, maybe a PUT to `/sessions` even though that's not exactly following the REST schema.

eteubert: Just saw that Phoenix has explicit support for what they call [Singleton Resources](https://hexdocs.pm/phoenix/1.4.8/Phoenix.Router.html#resources/4-singleton-resources). In their example for the user profile. But could be used for sessions as well.

eteubert:
Done implementing the Network REST actions. DIFF is unfortunately annoying to read now because of the renaming. Main file ist [lib/radiator_web/controllers/api/network_controller.ex](https://github.com/podlove/radiator/pull/137/files#diff-5b8a637d75b546af6da7be8864e224b7), which is the Controller equivalent to the previous GraphQL resolvers.

Off the top of my head guideline suggestions for writing REST controllers:

- don't check for user authentication as that is done automatically by the router pipeline
- always use `with` constructs so the FallbackController can pick up error tuples and render the appropriate response
- in general, controller code should only contain the "happy path", errors should be returned by the underlying business logic and turned to appropriate resonses by the FallbackController
- only use `create, show, update, delete` as public methods unless there is a good reason to deviate
- avoid resource nesting

guides/rest_api.md
monkeydom: The name here, since this is just the authentication point, the username/email should not be audio_file[name], but just name or another basename?

eteubert: 🤦‍♂  I took the tomatoes from my eyes, now I see it. Yes of course.

Issue #140, opened 2019-06-27
[User Story] User- & Permission-Management


eteubert:
**As a** network owner
**I want** to manage users and their permissions in my network

#### Acceptance Criteria

1. ~~Invite users to my network or any podcast / episode in my network.~~
2. See all users in my network and their permissions
3. Manage (add/remove) permissions for users in my network (except my own)

monkeydom:
Do you think 1. is a reasonable acceptance criteria? I would like to postpone that as that implies some sort of storage of uncommited data (e.g. I invite you, so you get an account creation email that in turn remembers that you also would have been given access to certain things)

eteubert:
No you're right, this can be moved to a separate story.

PR #142, opened 2019-06-28
User- & Permission-Management

monkeydom:
Early draft, but probably with worthwhile stuff to reason about.

Added a `Collaborator` struct to be used as interface to the `Editor` context.
```elixir
 @type t :: %__MODULE__{
          user: Auth.User.t(),
          subject: Network.t() | Podcast.t() | Episode.t() | Audio.t(),
          permission: permission()
        }
  defstruct [:user, :subject, :permission]
```

Editor grows `list_collaborators(actor, subject)` which eventually should return all users that have access to any part of the network, but currently only return network access having entities.

Editor also gains `add_collaborator(actor, %Collaborator{})

No Rest/GraphQL API yet.

Preliminary UI to put below the list in the Network part (later also in podcasts etc). No edit yet, just remove, you implicitly can edit by adding with a different permission. No autocomplete of usernames yet either.
Pasted_Image_28_06_19__12_47

Introduced a `SharedPartialsView` for reusable components. added `render_shared_partial(template, assings)` to our views for easy access.


monkeydom:
relates to #140

lib/radiator/directory/editor.ex
```diff
@@ -458,4 +458,89 @@ defmodule Radiator.Directory.Editor do
       @not_authorized_match
     end
   end
+
+  ## User Permission Management
+
+  # TODO: list all collaborators of underlying entities as well.
+
+  @spec list_collaborators(Auth.User.t(), Network.t()) ::
```
eteubert: If there's an `{:error, _}` reply, the happy path is usually `{:ok, _}`. Not 100% sure about the convention but I can't think of an example where that's not the case.

lib/radiator/directory/editor.ex
```diff
@@ -458,4 +458,89 @@ defmodule Radiator.Directory.Editor do
...
+    if has_permission(actor, subject, :manage) do
+      query =
+        from p in Ecto.assoc(subject, :permissions),
+          join: u in Auth.User,
+          on: p.user_id == u.id,
+          join: s in Network,
+          on: s.id == p.subject_id,
+          preload: [user: u]
+
+      with perm_list when is_list(perm_list) <- Repo.all(query) do
```
eteubert: The guard `is_list(perm_list)` is unnecessary. `Repo.all/1` always returns a list or raises. If you want to check for a nonempty list, I think the recommended way is `with perm_list = [_ | _] <- Repo.all(query)`

lib/radiator/directory/editor.ex
```diff
@@ -458,4 +458,89 @@ defmodule Radiator.Directory.Editor do
...
+  # TODO: Handle the accidental update case
+  @spec add_collaborator(Auth.User.t(), Collaborator.t()) ::
+          {:error, any} | {:ok, Collaborator.t()}
+  def add_collaborator(
+        actor = %Auth.User{},
+        collaborator = %Collaborator{user: user, subject: subject, permission: permission}
+      )
+      when is_permission(permission) do
+    if has_permission(actor, subject, :manage) do
+      case Editor.Permission.set_permission(user, subject, permission) do
```
eteubert: Use `with` to avoid `other -> other`? (applies to all 3 `add_collaborator` functions). 

```elixir
with :ok <- Editor.Permission.set_permission(user, subject, permission), do: {:ok, collaborator}
```

lib/radiator_web/controllers/admin/network_controller.ex
```diff
@@ -81,4 +90,66 @@ defmodule RadiatorWeb.Admin.NetworkController do
         other
     end
   end
+
+  def add_collaborator(conn, %{"id" => id, "collaborator" => collaborator}) do
```
eteubert: I try to keep controllers RESTful, even in CMS UIs. Here, the resource you're modifying is a "network collaboration". So I would have a `NetworkCollaborationController` (or generic `CollaborationController`) with `create` and `delete` implementation instead of `add_collaborator` and `remove_collaborator` here. It has the benefit of similar implementations/structure of CMS controllers and actual REST API controllers. It's encouraged by Phoenix through the [resources](https://hexdocs.pm/phoenix/1.4.8/Phoenix.Router.html#resources/4) Router Macro.

See also https://hexdocs.pm/phoenix/routing.html#resources

monkeydom: How would one adopt this for networks, podcasts and episodes alike? are these 3 different controller resources living under /:resourcepath/collaborator?

lib/radiator/directory/editor.ex
monkeydom: ok, can do. was a little ambigous of that because 
* it is nested already in an if (which we could also move to a more general expression used with a with?)
* i might want to handle some errors and then i'll end up back in this land. `other -> other` seems to be a bad pattern in general though. maybe something more suitably named? Also the internal submodules of Editor should probably not return something we passthrough directly anyways, so a match against a changeset we want to bubble up and other errors might be more appropriate here anyways

lib/radiator/directory/editor.ex
monkeydom: ah i see, was queuing from `Editor.list_podcasts` - although that doesn't handle the authentication yet either. So I take it we'll put everhting editor as {:ok, }. Need to audit a little probably.

lib/radiator/directory/editor.ex
monkeydom: k

lib/radiator_web/controllers/admin/network_controller.ex
monkeydom: I'll try. But I'm slightly confused why we still do have parallel controller hierarchies in API and UI then. E.g. should't that then just be switched based on the pipeline/resulttype vs. having parallel hierarchies?

eteubert: Mutations in particular are probably very similar in API and UI, true. Not aware of a way to bring these together but open to suggestions. Reading endpoints (index, show) though will often be different as the API has a narrow focus while the UI may display many different things at once.

To answer your first question, you can use resource nesting:

```elixir
resources "/networks", NetworkController do
  resources "/collaboration", CollaborationController, only: [:create, :delete]
end

resources "/podcasts", NetworkController do
  resources "/collaboration", CollaborationController, only: [:create, :delete]
end

# ...
```

That generates routes like `DELETE  /networks/:network_id/collaborations/:id`

In that case you can have one generic `CollaborationController` and match on the params, as one will have a `network_id` in the request and the other one `podcast_id`. etc. `mix phx.routes` is helpful to see what routes witch what paramters are generated.

lib/radiator/directory/editor.ex
eteubert: There is a pattern I have seen using `with` and "tagged" conditions that is a bit verbose but linearizes otherwise nested conditional if/case/with constructs:

```elixir
with {:permission, true} <- {:permission, has_permission(actor, subject, :manage)},
     {:set, :ok} <- {:set, Editor.Permission.set_permission(user, subject, permission)} do
  {:ok, collaborator}
else
  {:permission, _} -> @not_authorized_match
  {:set, _} -> raise {:error, :could_not_set_permission}
end
```

You code the "happy path" in the with but tag each line with a unique tuple. Then you can check in the `else` match which line failed and return the error accordingly.

lib/radiator/directory/editor.ex
```diff
@@ -458,4 +458,89 @@ defmodule Radiator.Directory.Editor do
...
+
+  # TODO: list all collaborators of underlying entities as well.
+
+  @spec list_collaborators(Auth.User.t(), Network.t()) ::
+          [
+            Collaborator.t()
+          ]
+          | {:error, any}
+  def list_collaborators(actor = %Auth.User{}, subject = %Network{}) do
+    if has_permission(actor, subject, :manage) do
```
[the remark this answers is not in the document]
eteubert: I see, go ahead 👍

lib/radiator/directory/editor.ex
monkeydom: ack

lib/radiator/directory/collaborator.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Radiator.Directory.Collaborator do
```
eteubert: I'm not too unhappy with the name `Collaborator` but my preferred naming is `Collaboration`. Reasoning: The user/person _is_ the collaborator but this entity describes the relationship between a user and subject: a collaboration.

lib/radiator_web/controllers/api/collaborator_controller.ex
```diff
@@ -0,0 +1,80 @@
...
+  use RadiatorWeb, :controller
+  use Radiator.Constants
+
+  alias Radiator.Directory.Editor
+
+  alias Radiator.Directory.Collaborator
+
+  action_fallback RadiatorWeb.Api.FallbackController
+
+  def create(conn, %{"username" => username, "permission" => permission} = params) do
```
eteubert: Feels too deeply nested to me. Here is a version that uses only one `with` block (for illustration, I didn't run it). However for finer grained control you would need the tagging as discussed before and that might become unreadable as well. The admin version of the controller is an even more interesting example as there are more failure handling cases already. Your call.

```elixir
  def create(conn, %{"username" => username, "permission" => permission} = params) do
    with actor <- current_user(conn),
         {:ok, subject} <- current_subject(actor, params),
         user when not is_nil(user) <- Radiator.Auth.Register.get_user_by_name(username),
         collaborator <- build_collaborator(user, permission, subject),
         {:ok, collaborator} <- Editor.add_collaborator(actor, collaborator) do
      conn
      |> put_status(:created)
      |> render("show.json", %{collaborator: collaborator})
    else
      _ -> {:error, :not_found}
    end
  end

  defp build_collaborator(user, permission, subject) do
    %Collaborator{
      user: user,
      permission: String.to_existing_atom(permission),
      subject: subject
    }
  end
```

lib/radiator_web/graphql/admin/resolvers/editor.ex
```diff
@@ -105,7 +105,13 @@ defmodule RadiatorWeb.GraphQL.Admin.Resolvers.Editor do
   end
 
   def find_user(_, _, %{context: %{current_user: user}}) do
-    {:ok, user}
+    user
+    |> Radiator.Repo.preload(:person)
+    |> (&{:ok, &1}).()
```
eteubert: Might be worth considering adding a helper function that wraps anything in an :ok tuple for readability. If we can come up with a good name.

lib/radiator/auth/register.ex
```diff
@@ -39,6 +41,16 @@ defmodule Radiator.Auth.Register do
     end
   end
 
+  def find_users(query_string \\ "") do
```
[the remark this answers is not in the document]
monkeydom: I think my intent is eventually to disallow this case altogether ^^. Managing users in general is probably a separate instance-admin-api. But since I forgot to sort them I think I should do another case here for now.

lib/radiator/directory/collaborator.ex
monkeydom: To me collaboration would be work product. E.g. the result of this, or the ongoing activity. So I would leave it like this for now since you are not too unhappy. ^^

lib/radiator_web/graphql/admin/resolvers/editor.ex
monkeydom: Yeah, Like Tuple.ok(). Kind of dislike the ugliness of this statement here, but like that it is concise.

guides/rest_api.md
```diff
@@ -1,29 +1,42 @@
...
-  - [Read](#Read-1)
-  - [Update](#Update-1)
-  - [Delete](#Delete-1)
-- [Episodes](#Episodes)
-  - [Parameters for Create & Update](#Parameters-for-Create--Update-2)
-  - [Create](#Create-2)
-  - [Read](#Read-2)
-  - [Update](#Update-2)
-  - [Delete](#Delete-2)
+- [REST API](#REST-API)
```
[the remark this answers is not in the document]
monkeydom: ok. Also the current structure looks quite ugly in the actual generated docs.

config/config.exs
```diff
@@ -80,6 +81,11 @@ config :logger, :console,
 # Use Jason for JSON parsing in Phoenix
 config :phoenix, :json_library, Jason
 
+config :ex_aws,
```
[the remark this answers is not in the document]
monkeydom: yeah - my bad.

config/config.exs
```diff
@@ -8,6 +8,7 @@
 use Mix.Config
 
 default_host = "localhost"
+default_host = "darthdom.local"
```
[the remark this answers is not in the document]
monkeydom: 👍

lib/radiator_web/controllers/api/collaborator_controller.ex
monkeydom: Well, you are killing the not authorized errors though. Get the point, Took the idea and reworked it a bit.

PR #145, opened 2019-06-30
Elixir 1.9 Releases

eteubert:
- use Elixir 1.9 Releases to build binary release
- optimize Dockerfile by using multi stage builds
- replace mix ecto.create/migrate with our own implementation in `Radiator.Release` as Mix is not available in releases
- extract all runtime configs to environment variables so they


eteubert:
Looks like everything is working. Once this is merged I can look into continuous deployment for a staging environment for us.

PR #146, opened 2019-07-02
add: permission check for multiple possible parents

eteubert:


lib/radiator/directory/editor/permission.ex
```diff
@@ -51,43 +51,49 @@ defmodule Radiator.Directory.Editor.Permission do
...
-    |> Repo.one!()
+  defp parents(subject = %Audio{}) do
+    network =
+      subject
+      |> Ecto.assoc(:network)
+      |> Repo.one()
+
+    episodes = subject |> Ecto.assoc(:episodes) |> Repo.all()
+
+    [network | episodes]
```
monkeydom: hmmm, so in your implementation this can return nil in the list. I don't think this is a good contract. Would only add network if not nil.

eteubert:
Great feedback, that simplifies the logic and makes it easier to read 💯 

Can you think of a more concise way to handle network being nil than this?

```elixir
  case network do
    nil -> episodes
    network -> [network | episodes]
  end
```

monkeydom:
```elixir
cond do
  network -> [network | episodes]
  true -> episodes
end
```

```elixir
if network do
  [network | episodes]
else
  episodes
end
```

```elixir
if network, do: [network | episodes], else: episodes
```

all not really an improvement, maybe the online if though.

eteubert:
Seeing it all together I like the `if` because it doesn't need the `nil` or `true` in the code which makes it look cleaner. I know it's only hidden through the `else` but ¯\\_(ツ)_/¯

PR #148, opened 2019-07-03
REST API: add Audio and AudioFile support

eteubert:


guides/rest_api.md
```diff
@@ -257,3 +289,62 @@ PUT /api/rest/v1/episodes/:id
...
+POST /api/rest/v1/audios
+```
+
+### Read
+
+```
+GET /api/rest/v1/audios/:id
+```
+
+### Update
```
monkeydom: As far as I read the rest stuff: Put should be used only if the entity is replaced completely, for field by field changes PATCH should be used. Since we rarely require the whole update, should we reference PATCH for all our changes instead of PUT?

lib/radiator/directory/editor/manager.ex
```diff
@@ -67,6 +88,10 @@ defmodule Radiator.Directory.Editor.Manager do
     |> Repo.update()
   end
 
+  def delete_audio(%Audio{} = audio) do
+    Repo.delete(audio)
```
monkeydom: does this cascade correctly? e.g. is the audio removed from all entities it is associated with and does it delete an audio_file if it is the only user?

lib/radiator_web/controllers/api/audio_controller.ex
```diff
@@ -0,0 +1,59 @@
...
+         {:ok, audio} <- Editor.update_audio(user, audio, audio_params) do
+      render(conn, "show.json", %{audio: audio})
+    end
+  end
+
+  def delete(conn, %{"id" => id}) do
+    with user <- current_user(conn),
+         {:ok, audio} <- Editor.get_audio(user, id),
+         {:ok, _} <- Editor.delete_audio(user, audio) do
+      send_resp(conn, 204, "")
```
monkeydom: isn't the convention to return the deleted entity?

lib/radiator/directory/editor/manager.ex
eteubert: associated `audio_file`s are deleted via Postgres constraints set in the migration. I wasn't sure so I added a test to verify.

It also deletes the associated episode, which feels to destructive. I'm going to change it from `:delete_all` to `:nilify_all` there.

lib/radiator_web/controllers/api/audio_controller.ex
eteubert: Not sure, couldn't find a definitive answer. The most "proper" excerpt I can offer is from https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.7 about the DELETE verb in general:

> A successful response SHOULD be 200 (OK) if the response includes an entity describing the status, 202 (Accepted) if the action has not yet been enacted, or 204 (No Content) if the action has been enacted but the response does not include an entity. 

I also found arguments that DELETE should return a 20x instead of 404 for nonexistent entities to make it idempotent. Then 204 is the only option.

In any case, we should write down how we treat DELETE in general and put it in the docs. Not sure if it's useful for UIs to get the deleted object back in the response (the UI should already have the object?) but the main concern is consistency.

PR #152, opened 2019-07-04
Issue/101 REST API: Chapter Management

### LINKED ISSUE #101: [User Story] Chapter Editing with REST API
**As a** API client
**I want** to be able to query, update and add chapters of podcast episodes

#### Acceptance Criteria

1. ability to upload an image using post data
1. ~~ability to query chapters of an episode, including all attributes (`start, title, link, image_url`)~~
1. ability to change a chapter 
1. ability to delete a chapter

eteubert:
1) Calling a plug directly in https://github.com/podlove/radiator/compare/issue/101-api-chapter-management?expand=1#diff-d0559d752a76cec019da187d4034d879R43 feels strange but the general idea of short-circuiting here should be ok. 

2) I added

```elixir
def get_permission(_user = %Auth.User{}, _subject = %Chapter{}), do: nil
```

so the permission check on chapter always fails and falls back to the parent (Audio). Is that the intended place? Should there be a general `def get_permission(_user = %Auth.User{}, _subject), do: nil` catchall?

closes #101


monkeydom:
Yeah - I think probably for short circuiting errors a `raise` with a corrsponding catch in the error view is probably more appropriate.

lib/radiator/directory/editor/permission.ex
```diff
@@ -25,6 +26,9 @@ defmodule Radiator.Directory.Editor.Permission do
   def get_permission(user = %Auth.User{}, subject = %Audio{}),
     do: do_get_permission(user, subject)
 
+  def get_permission(_user = %Auth.User{}, _subject = %Chapter{}),
+    do: nil
+
```
monkeydom: Hmm - this is kind of wonky. Since we don't have any permissions on Chapters (and we don't want to have them), I think it is more appropriate to reference the audio in the callsite, rather than do this monkeypatching here.

E.g. all callsites should go:
```elixir
has_permission(actor. %Audio{id: chapter.audio_id}, :edit)
```

instead. If you insist to expose the Chapter here, then make it a short circuit of the has_permission
```elixir
def has_permission(actor = %Auth.User{}, _subject = %Chapter{audio_id: audio_id},permission), do: has_permission(actor, %Audio{id: audio_id}, permission)
```

lib/radiator/directory/editor/permission.ex
```diff
@@ -83,6 +87,13 @@ defmodule Radiator.Directory.Editor.Permission do
     |> List.wrap()
   end
 
+  defp parents(subject = %Chapter{}) do
+    subject
+    |> Ecto.assoc(:audio)
+    |> Repo.one!()
+    |> List.wrap()
+  end
+
```
monkeydom: if you heed my advice, this can be removed again.

lib/radiator_web/views/api/chapters_view.ex
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def render("chapter.json", %{conn: conn, chapter: chapter, audio: audio}) do
+    %Document{}
+    |> Document.add_link(%Link{
+      rel: "self",
+      href: Routes.api_audio_chapters_path(conn, :show, audio.id, chapter.id)
+    })
+    |> Document.add_properties(%{
+      id: chapter.id,
```
monkeydom: imho this needs at least the `audio_id` as well to be useful for e.g. update, as we need the audio id in the update path too

lib/radiator/directory/editor.ex
```diff
@@ -421,6 +423,44 @@ defmodule Radiator.Directory.Editor do
     episode
   end
 
+  def get_chapter(actor = %Auth.User{}, id) do
```
monkeydom: I'd rather have us use the combination audio_id,start as primary keys here. of course multiple identical start values are forbidden then and need to be constraint against.

lib/radiator/directory/editor.ex
```diff
@@ -421,6 +423,44 @@ defmodule Radiator.Directory.Editor do
...
+    if has_permission(actor, audio, :edit) do
+      AudioMeta.create_chapter(audio, attrs)
+    else
+      @not_authorized_match
+    end
+  end
+
+  def update_chapter(actor = %Auth.User{}, chapter = %Chapter{}, attrs) do
+    if has_permission(actor, chapter, :edit) do
+      AudioMeta.update_chapter(chapter, attrs)
```
monkeydom: update_chapter needs to delete the old image file if a new one is uploaded. otherwise we leak files in the store (just tried by updating some chapter)

lib/radiator/directory/editor.ex
```diff
@@ -421,6 +423,44 @@ defmodule Radiator.Directory.Editor do
...
+    else
+      @not_authorized_match
+    end
+  end
+
+  def delete_chapter(actor = %Auth.User{}, chapter = %Chapter{}) do
+    if has_permission(actor, chapter, :own) do
+      AudioMeta.delete_chapter(chapter)
```
monkeydom: delete chapter also needs to remove the file (probably also directory) from s3. currently leaks.

lib/radiator_web/controllers/api/chapters_controller.ex
```diff
@@ -0,0 +1,58 @@
...
+  defp assign_audio(conn, _) do
+    conn
+    |> current_user()
+    |> Editor.get_audio(conn.params["audio_id"])
+    |> case do
+      {:ok, audio} ->
+        conn
+        |> assign(:audio, audio)
+
+      error = {:error, _} ->
```
monkeydom: As mentioned before, this is probably better done using a raise and catch the exception the way the original not found is done

lib/radiator_web/controllers/api/chapters_controller.ex
```diff
@@ -0,0 +1,58 @@
...
+  def update(conn, %{"id" => id, "chapter" => chapter_params}) do
+    with user <- current_user(conn),
+         {:ok, chapter} <- Editor.get_chapter(user, id),
+         {:ok, chapter} <- Editor.update_chapter(user, chapter, chapter_params) do
+      conn
+      |> assign(:chapter, chapter)
+      |> render("show.json")
+    end
+  end
+
```
monkeydom: missing `:show`

eteubert:
Ready for a second review. A few notes:

1. Regarding calling `RadiatorWeb.Api.FallbackController` directly in `ChaptersController`: I don't think we can raise in a plug as it breaks the typespec contract and Dialyzer complains. Tried researching a bit for similar implementations but couldn't find any, so I left it as it is for now.
2. Leaking files in arc: I am surprised arc_ecto doesn't handle this automatically. My current solution for chapter images works "good enough" as far as my testing goes but our aim should be a generic changeset step that we can throw in everywhere we use an arc_ecto attribute. See #166

lib/radiator_web/controllers/api/chapters_controller.ex
```diff
@@ -3,7 +3,12 @@ defmodule RadiatorWeb.Api.ChaptersController do
 
   alias Radiator.Directory.Editor
 
-  plug :assign_audio when action in [:create, :update]
+  plug :assign_audio when action in [:show, :create, :update, :delete]
+  plug :assign_chapter when action in [:show, :update, :delete]
```
monkeydom: ah - i see what you did there. Interesting pattern.

lib/radiator_web/controllers/api/chapters_controller.ex
```diff
@@ -55,4 +58,17 @@ defmodule RadiatorWeb.Api.ChaptersController do
...
+
+  defp assign_chapter(conn, _) do
+    with {:ok, chapter} <-
+           Editor.get_chapter(current_user(conn), conn.assigns[:audio], conn.params["start"]) do
+      conn
+      |> assign(:chapter, chapter)
+    else
+      error = {:error, _} ->
+        conn
+        |> RadiatorWeb.Api.FallbackController.call(error)
```
monkeydom: Hmm, shouldn't we somehow just refer to the assigned fallbackcontroller? e.g. something like 
```elixir
apply(@phoenix_fallback, :call, [error])
```
? - in the middle of some other change now, can't verify if that is the way this should work. but the falback seems to be in the phoenix_fallback attribute

eteubert:
Thanks for pointing out `@phoenix_fallback`. However that's a tuple, either `{:module, mod}` or `{:function, fun}`, depending on type of plug used.

What I came up with:

```elixir
  defp apply_action_fallback(conn, response) do
    case @phoenix_fallback do
      {:module, module} -> apply(module, :call, [conn, response]) |> halt()
    end
  end
```

Then I can 

```elixir
  defp assign_chapter(conn, _) do
    with {:ok, chapter} <-
           conn
           |> current_user()
           |> Editor.get_chapter(conn.assigns[:audio], conn.params["start"]) do
      conn
      |> assign(:chapter, chapter)
    else
      response -> apply_action_fallback(conn, response)
    end
  end
```

It's at least removes the direct reference to the plug module name, which is nice. What bothers me is that I cannot add the `{:function, fun}` case because Dialyzer determines it's a module and complains the match is pointless. Anyway, still happier with this than before.

PR #160, opened 2019-07-12
add(api/graphql): query an episode by guid

### LINKED ISSUE #155: GraphQL: query episodes by guid
requested by @alexander-heimbuch for „Preview“

eteubert:
fixes #155

### DIFF
--- a/lib/radiator_web/graphql/admin/resolvers/editor.ex
+++ b/lib/radiator_web/graphql/admin/resolvers/editor.ex
@@ -72,7 +72,14 @@ defmodule RadiatorWeb.GraphQL.Admin.Resolvers.Editor do
   """
   defmacro with_episode(user, episode_id, do: block) do
     quote do
-      case Editor.get_episode(unquote(user), unquote(episode_id)) do
+      with_episode(unquote(user), unquote(episode_id), :get_episode, do: unquote(block))
+    end
+  end
+
+  defmacro with_episode(user, episode_id, method, do: block) do
+    quote do
+      apply(Editor, unquote(method), [unquote(user), unquote(episode_id)])
+      |> case do
         {:ok, episode} -> unquote(block).(episode)
         @not_found_match -> @not_found_response
         @not_authorized_match -> @not_authorized_response
@@ -218,6 +225,16 @@ defmodule RadiatorWeb.GraphQL.Admin.Resolvers.Editor do
   def find_episode(_parent, %{id: id}, %{context: %{current_user: user}}) do
     with_episode user, id do
       fn episode -> {:ok, episode} end
+    end
+  end
+
+  def find_episode(_parent, %{guid: guid}, %{context: %{current_user: user}}) do
+    try do
+      with_episode user, guid, :get_episode_by_guid do
+        fn episode -> {:ok, episode} end
+      end
+    rescue
+      Ecto.MultipleResultsError -> {:error, "There is more than one episode with given guid"}
     end
   end
 
--- a/lib/radiator_web/graphql/admin/resolvers/editor.ex
+++ b/lib/radiator_web/graphql/admin/resolvers/editor.ex
@@ -69,6 +69,10 @@ defmodule RadiatorWeb.GraphQL.Admin.Resolvers.Editor do
         fn episode -> {:ok, episode}
       end
 
+      with_episode user, guid, :get_episode_by_guid do
+        fn episode -> {:ok, episode} end
+      end
+
   """
   defmacro with_episode(user, episode_id, do: block) do
     quote do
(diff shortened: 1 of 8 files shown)


lib/radiator_web/graphql/admin/resolvers/editor.ex
```diff
@@ -221,6 +228,16 @@ defmodule RadiatorWeb.GraphQL.Admin.Resolvers.Editor do
     end
   end
 
+  def find_episode(_parent, %{guid: guid}, %{context: %{current_user: user}}) do
+    try do
+      with_episode user, guid, :get_episode_by_guid do
+        fn episode -> {:ok, episode} end
+      end
+    rescue
+      Ecto.MultipleResultsError -> {:error, "There is more than one episode with given guid"}
```
monkeydom: See above. I'd rather narrow the list down to the ones the user has access to first. And then return. Also not sure if alex didn't want a public api here instead.

lib/radiator_web/graphql/schema.ex
```diff
@@ -66,7 +66,8 @@ defmodule RadiatorWeb.GraphQL.Schema do
 
     @desc "Get one published episode"
     field :published_episode, :published_episode do
-      arg :id, non_null(:id)
+      arg :id, :id
+      arg :guid, :string
```
monkeydom: I think I would prefer a more open endend query api, that also has guid as optional query element and returns a list instead.

test/radiator_web/graphql/admin/schema/query/episodes_test.exs
```diff
@@ -97,6 +97,51 @@ defmodule RadiatorWeb.GraphQL.Admin.Schema.Query.EpisodesTest do
...
+               "episode" => %{
+                 "id" => Integer.to_string(episode.id),
+                 "title" => episode.title,
+                 "guid" => "guid-foo"
+               }
+             }
+           }
+  end
+
+  test "episode by guid returns an episode when non existent", %{conn: conn} do
```
monkeydom: shouldn't the test names be positive? E.g. `returns not found when non-existent` (and apparently, non-existent is hyphenated according to the macOS dictionary)

test/radiator_web/graphql/admin/schema/query/episodes_test.exs
```diff
@@ -97,6 +97,51 @@ defmodule RadiatorWeb.GraphQL.Admin.Schema.Query.EpisodesTest do
...
+    conn = get conn, "/api/graphql", query: @single_guid_query, variables: %{"guid" => "guid-foo"}
+
+    assert %{"errors" => [%{"message" => message}]} = json_response(conn, 200)
+    assert message =~ "not found"
+  end
+
+  test "episode by guid returns an episode when there is more than one episode with that guid", %{
```
monkeydom: dito above.

lib/radiator_web/graphql/admin/resolvers/editor.ex
```diff
@@ -72,7 +72,14 @@ defmodule RadiatorWeb.GraphQL.Admin.Resolvers.Editor do
   """
   defmacro with_episode(user, episode_id, do: block) do
     quote do
-      case Editor.get_episode(unquote(user), unquote(episode_id)) do
+      with_episode(unquote(user), unquote(episode_id), :get_episode, do: unquote(block))
+    end
+  end
+
+  defmacro with_episode(user, episode_id, method, do: block) do
```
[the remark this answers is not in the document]
eteubert: Renamed it and added an example to the macro doc. Maybe we can revert this whole addition if we go with guid expecting more than one result.

eteubert:
@alexander-heimbuch can you comment on how you expect to use this? Public or Authed API? What would you expect as a result if there are multiple episodes with the same guid in the system?

eteubert:
@monkeydom Since we can't guarantee guids being what their name suggests, globally unique, your proposal to return a list sounds good to me. I would then move the `guid` arg from `:episode` to `:episodes` field (which apparently doesn't even exist yet) and use it as a filter argument there.

monkeydom:
> @monkeydom Since we can't guarantee guids being what their name suggests, globally unique, your proposal to return a list sounds good to me. I would then move the `guid` arg from `:episode` to `:episodes` field (which apparently doesn't even exist yet) and use it as a filter argument there.

Yes, i'd like to see that. Top level query, `publishedEpisodes` and `episodes` with some query args.

alexander-heimbuch:
A `guid` in my specific case is useful if it is a unique identifier. Otherwise I would simply stick to the `id`. Still a more opened query API would very useful for us. For example searching the title/description and also sorting the results.

monkeydom:
Let's shelf this for now.

PR #162, opened 2019-07-17
feature(api, content): document and cleanup data structures and apis, partial: network and podcast

### LINKED ISSUE #161: Content: document data fields and their meaning/implication for podcast/episodes

monkeydom:
Documents the intentional change, converts and fixes along the way.

Network: add image to api, make `publishedNetworks` only show published networks
Podcast: rename description to summary, use consistently, add and expose more fields in apis

needs ecto reset
this fixes #161 partially


guides/data_and_structure.md
```diff
@@ -0,0 +1,151 @@
...
+
+Potential fields open for discussion:  
+
+* `link`
+* More iTunes metadata fields  (`itunes:explicit`, `itunes:category`)
+* Contributors
+
+Existing fields, worthy of discussion:
+
+* `publishedAt` - is currently both put into the feeds `pubDate` tag (which is quite unused by clients/directories) and used as a switch to say if a podcast is public. @monkeydom: the removing the `publishedAt` would be prudent, as it is ambiguous, and the mechanism to determine the published state of a podcast should be handled differently.
```
eteubert: No harm in filling the `pubDate` field thoug? But I agree, the publish state should be handled separately.

guides/data_and_structure.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Episode 
+
+A **Episode** belongs to exactly a **Podcast**
+
+ Field | Type | Description 
+ ----: | :--  | :--
+`shortID` | `String` | Full short ID of the episode. Defaults to `<shortIDOfPodcast><episodenumberWithLeadingZeros>`, e.g. `FS002`.
```
eteubert: `episodenumberWithLeadingZeros`: Any opinion on how many digits by default? 3?

guides/data_and_structure.md
```diff
@@ -0,0 +1,151 @@
...
+ `title` | `String` | The name of the episode. Will be put into `title`, and `itunes:title`. If prefixed with the short ID, the shortID and following whitespace will be stripped for the `itunes:title` to conform with Apple's guidelines there
```
eteubert: Uhh, magic. Must be configurable at least. If the episode `shortID` is an important communication device for the podcaster, he/she must have the option to keep it in the title. Because if you follow Apple's guidelines the shortID is not displayed anywhere in podcast clients. //cc @timpritlove has opinions

guides/data_and_structure.md
```diff
@@ -0,0 +1,151 @@
...
+ 
+ 
+Potential fields open for discussion: 
+
+* `slug` 
+* `link`
+
+### Podcast
+
+A **Podcast** belongs to exactly a **Network**
```
eteubert: Might be a good idea to define exact wording to describe relationships. "_a/an_ entity" is ambiguous. Suggestions:

- X belongs to exactly one Y (1:1, NULL not allowed)
- X belongs to no Y or exactly one Y (1:1, NULL allowed)
- X belongs to at least one Y (1:n, NULL not allowed)
- X belongs to no Y or any number of Y (1:n, NULL allowed)

guides/data_and_structure.md
```diff
@@ -0,0 +1,151 @@
...
+
+A **Contribution** references a Person and either an **Audio** or a **Podcast**
+
+ Field | Type | Description 
+   --: | :--  | :--
+ `role` | `String` | A contribution role for that person. E.g. `on_air` or `support`
+ 
+ Discussion:
+ 
+ * How much should we formalize this? Currently this is just a text string. I think we should at least make a list of suggestions for the default roles we imagine, that can be offered in frontends and given a good correspondence in the feed metadata too.
```
eteubert: Providing a predefinet set of roles sounds like a worthwhile goal. Reduces complexity and makes the feature more approachable.

Maybe @timpritlove can chime in with his experience of working with Contributor roles & groups in Podlove Publisher.

lib/radiator/directory.ex
```diff
@@ -30,9 +30,19 @@ defmodule Radiator.Directory do
...
+  defp published_networks_query do
+    published_podcasts_query = PodcastQuery.filter_by_published(Podcast)
+
+    network_query =
+      from(p in published_podcasts_query, join: n in assoc(p, :network), distinct: n, select: n)
+
+    from(n in subquery(network_query), order_by: [desc: n.title])
+  end
+
+  # TODO: have a clearer concept of published for networks (currently just defers to the podcast)
```
eteubert: I've been thinking about this before and I wonder if we can get away without a network publish state. Your deferrence to podcasts seems fine. Do we ever need a network be public when there is no public podcast inside?

guides/data_and_structure.md
monkeydom: Went through the specs, didn't find any reference to it. I think it is ambiguous, and at best should just default to the first episode ever published probably. Nobody is using the data, so I think removing it is nicer, just unused unclear semantics otherwise.

guides/data_and_structure.md
monkeydom: I would go for three, yes. Eventually probably a sort of customizeable pattern generation should be put in place imho.

guides/data_and_structure.md
monkeydom: okay, will work that in. Probably should also be put on both sides. currently I opted for just listing the belongs-to sides.

lib/radiator/directory.ex
monkeydom: Only in terms of a pre-announcement. Which usually is done by publishing an 0-episode, so all of this might be fine.

guides/data_and_structure.md
eteubert: Using publishing date of first episode sounds good. My +1 for removing it then :)

PR #169, opened 2019-07-18
Refine chapters

eteubert:
- add chapters index
- don't include chapters in episode


lib/radiator_web/views/api/chapters_view.ex
```diff
@@ -1,14 +1,22 @@
...
+      embed: render_many(assigns[:chapters], __MODULE__, "chapter.json", assigns)
+    })
+  end
+
   def render("show.json", assigns) do
     render(__MODULE__, "chapter.json", assigns)
   end
 
-  def render("chapter.json", %{conn: conn, chapter: chapter, audio: audio}) do
+  def render("chapter.json", %{conn: conn, chapters: chapter, audio: audio}) do
```
monkeydom: why the key chapters if you only render one?, shouldn't it still be chapter?

lib/radiator_web/views/api/chapters_view.ex
```diff
@@ -1,14 +1,22 @@
...
   use RadiatorWeb, :view
 
-  alias HAL.{Document, Link}
+  alias HAL.{Document, Link, Embed}
   alias Radiator.AudioMeta.Chapter
 
+  def render("index.json", assigns) do
+    %Document{}
+    |> Document.add_embed(%Embed{
+      resource: "rad:chapter",
```
monkeydom: haven't looked at the result. But wouldn't it be rad:chapters as container? I'm unclear what a top level container should look like here though.

lib/radiator_web/views/api/chapters_view.ex
eteubert: Yep, fixed already. The `render_many` put the chapter under the `chapters` key for some reason, so I rewrote it with a manual `Enum.map` and now I can use the `chapter` key again.

lib/radiator_web/views/api/chapters_view.ex
eteubert: I think the resource name is always supposed to be singular. The example in http://stateless.co/hal_specification.html includes `ea:order`, which is a list.

eteubert:
Oh, totally forgot about this PR. I cherry-picked 31b3f94 into #194, the rest seems obsolete now.

Issue #171, opened 2019-07-22
API: provide concept of a task in API and Phoneix-API to be used for long running tasks like import 


monkeydom:
provide a minimal implementation to be used and expanded upon

Issue #25, opened 2019-04-04
Add API to import rss feeds


monkeydom:
Related to #1

monkeydom:
Needs to enable [radiator-cms #60: As podcaster I want to import podcasts](https://github.com/podlove/radiator-cms/issues/60)

Multistep API to be designed. In its simplest form it needs to take an URL and present a potential import candidate behind this api, then an API to trigger the import with some options, and then finally report progress on that import and also provide some errors along the way if any.

PR #173, opened 2019-07-25
feat(api): import api by creating a task api

monkeydom:
Tasks can be started via 

`POST /api/rest/v1/tasks`

currently only the import podcast feed task exists:

 Name                                   | Type                | Description                                                                                                                                            |
| -------------------------------------- | ------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
| `import_podcast_feed[network_id]`      | `integer`           | Network ID. (required).                                                                                                                                |
| `import_podcast_feed[feed_url]`        | `string`            | feed or show url (check availablity and extent with GraphQL query `feedInfo`                                                                           |
| `import_podcast_feed[enclosure_types]` | `array` of `string` | Optional. E.g. `["audio/mpeg","audio/mp4"]` or [] for none. Defaults to all available enclosures (not implemented yet, currently only mp3 is imported) |
| `import_podcast_feed[short_id]`        | `string`            | Short ID for the podcast. E.g. `"CRE"`, `"FAN"`, `"FG"`.                                                                                               |
| `import_podcast_feed[limit]`           | `integer`           | Limits the amount of episodes to import. Optional, Defaults to all found.                                                                              |

The status of that task can be read using:

```
GET /api/rest/v1/tasks/:id
```

Example:
```json
{
  "end_time": "2019-07-25T12:18:40.810667Z",
  "id": 543,
  "progress": 4,
  "start_time": "2019-07-25T10:55:38.521178Z",
  "state": "exited",
  "title": "Import 'freakshow.fm' into ACME",
  "total": 8,
  "_links": {
    "rad:subject": {
      "href": "/api/rest/v1/podcasts/14"
    },
    "self": {
      "href": "/api/rest/v1/tasks/543"
    }
  }
}
```

| Name         | Type                                         | Description                                      |
| ------------ | -------------------------------------------- | ------------------------------------------------ |
| `state`      | one of `setup`,`running`,`finished`,`exited` |
| `progress`   | `integer`                                    | amount of items finished                         |
| `total`      | `integer`                                    | total amount of items                            |
| `title`      | `string`                                     | Human readable english title describing the task |
| `start_time` | `timestamp`                                  |
| `end_time`   | `timestamp`                                  | can be nil. time the task `finished` or `exited` |

Currently needs polling, in the future one might add a channel/subscription


monkeydom:
fixes #171 and #25

eteubert:
When state is `exited`, should we add the ability to store one or many error descriptions?

lib/radiator/directory/episode.ex
```diff
@@ -143,6 +143,18 @@ defmodule Radiator.Directory.Episode do
...
+  def generate_short_id(short_id_base, number) when is_integer(number) do
+    generate_short_id(short_id_base, to_string(number))
+  end
+
+  def generate_short_id(short_id_base, number) when is_binary(number) do
+    "#{short_id_base}#{String.pad_leading(number, 3, "0")}"
+  end
+
+  def generate_short_id(short_id_base, _) do
+    generate_short_id("#{short_id_base}_t", :rand.uniform(1000))
```
eteubert: What's the reasoning here to generate a random short id instead of leaving it blank?

lib/radiator_web/controllers/api/task_controller.ex
```diff
@@ -0,0 +1,60 @@
...
+
+  alias Radiator.Task.TaskManager
+
+  alias Radiator.Directory.{
+    Editor,
+    Importer
+  }
+
+  def create(conn, %{"import_podcast_feed" => import_feed_params}) do
+    import_opts =
```
eteubert: I like to use Ecto for controller param whitelisting, validation and casting. Keeps the controller leaner / focused. I used to create a module for each but I think it's now possible inline with [schemaless changesets](https://hexdocs.pm/ecto/Ecto.Changeset.html?source=post_page---------------------------#module-schemaless-changesets).

But your way works fine as well.

monkeydom:
I think so. However, I wanted to keep the structure minimal for now (e.g. also not provide any mechanism to describe the unit, etc.). I think probably a good thing would be a whole progress/error log to provide eventually on exit.

lib/radiator/directory/episode.ex
monkeydom: Not much, more the idea if we want short_ids to provide some independent from the fact if the imported one had episode numbers. And since we can't reliably deduce episode numbers going random was just were I got to.

lib/radiator_web/controllers/api/task_controller.ex
monkeydom: Interesting. Maybe worth a look.

monkeydom:
Yeah - oban or another job solution is probably better. I didn't find a clear thing to add so I rolled a basic one on my own. I'm happy to adopt something else soonish.

Issue #139, opened 2019-06-27
People/Contributions: REST API


eteubert:
Add ability to manage people and their contributions to podcasts and audios via REST API.

PR #188, opened 2019-07-28
feature(api): rest api for people

monkeydom:
fixes part of #139 

also happening: switched most `:string` types to `:text` in migrations. Removed UserAvatar as it is unused. Made person avatar and podcast avatar work on first upload (was incorrectly splitting up inserts). Other renames on person to be more consistent (`link`,`image`)

Also updated the documentation files with weekend infos.

graphql and contributions to go.


lib/radiator/directory/editor/editor.ex
```diff
@@ -0,0 +1,52 @@
...
+  """
+  import Ecto.Query, warn: false
+
+  alias Ecto.Multi
+
+  alias Radiator.Repo
+
+  alias Radiator.Directory.{
+    Network
+  }
```
eteubert: Only use `{}` syntax when there are at least two entries? Otherwise it looks too verbose.

lib/radiator/directory/editor/editor.ex
```diff
@@ -0,0 +1,52 @@
+defmodule Radiator.Directory.Editor.Editor do
```
eteubert: `Editor.Editor` is very confusing. I understand the naming -- both editors are actually different -- but it feels time to come up with a  better name for `Radiator.Directory.Editor` as _this_ module is appropriately named `Editor`.

I've seen `CMS` used for something like this somewhere but it opens up another confustion for us because of the already existing `radiator-cms`. `Radiator.Directory.EditorAsInTheSoftware.EditorAsInThePerson` ¯\\_(ツ)_/¯

lib/radiator_web/controllers/api/person_controller.ex
```diff
@@ -0,0 +1,59 @@
+defmodule RadiatorWeb.Api.PersonController do
+  use RadiatorWeb, :rest_controller
+
+  alias Radiator.Directory.Editor
+  alias Radiator.Directory.Network
+
+  require Logger
+
+  # TODO: move this into :rest_controller for all rest controllers
+  def action(conn, _) do
```
eteubert: what does this do?

monkeydom: add the user to the action function calls as param, vs. having to fetch it in every action with current_user. And as all our REST api needs a user it seems appropriate. It is also a style used in the Phoenix book.

Issue #178, opened 2019-07-27
REST API: Clean up audio & published states


eteubert:
An audio is _either_ in a network _or_ in one/many episodes, this is enforced on creation and cannot be changed afterwards.

An unpublished audio:
- cannot be viewed
- cannot be downloaded

Published state in the audio is only used for network audios.

Episode audio published state is equal to the episode published state (audio published state is ignored in this case).

eteubert:
See also  #180

PR #189, opened 2019-07-30
#180 introduce AudioPublication

### LINKED ISSUE #180: Introduce AudioPublication
A layer between Audio and Network that takes publication state out of the Audio.

eteubert:
- add entity `AudioPublication` between Network and Audio
- remove Audio `published_at` attribute
- permissions: `Audio` does not have permissions any more; add `AudioPublication` instead
- stub `Directory.is_published` for now as this will soon be rewritten with state mechanics
- stub tracking `hours_since_published`, will fix it when I pick up the tracking issue

Redoing the publishing logic is a separate issue (#178)

fixes #180


lib/radiator/directory/editor.ex
```diff
@@ -519,21 +519,45 @@ defmodule Radiator.Directory.Editor do
...
   end
 
+  # thought: we need this logic somewhere, shouldn't it be in the
+  # permission context as the general logic is already there?
+  # I feel like it makes more sense bringing objects without permissions,
+  # like Audio, into the permissions module, than handling permission
+  # logic here.
+  # But that would mean supporting has_permission/3 on subjects
+  # that do not actually have their own permissions.
+  defp can_access_audio(actor, audio, permission) do
```
eteubert: One the other hand, maybe it's fine keeping this locally as long as it's only used from withing this Editor module? I keep going back and forth on this one 🤔

monkeydom: I like it separate as long as we don't want to have it abstracted. IMHO it is good that in permissions only the entities with actuall permissions are handeled. Any derived permissions living farther out sounds good to me in terms nobody gets confused if audio permissions are separate or not. And after all, @timpritlove isn't satisfied with the current permission model anyway, so this will probably end up being a bigger and different task anyways.

lib/radiator/directory/editor.ex
```diff
@@ -519,21 +519,45 @@ defmodule Radiator.Directory.Editor do
...
+    Enum.any?(get_audio_parents(audio), fn parent ->
+      has_permission(actor, parent, permission)
+    end)
+  end
+
+  defp get_audio_parents(audio) do
+    network =
```
monkeydom: imho this should be `audio_publication`

lib/radiator/directory/editor/manager.ex
```diff
@@ -84,9 +84,19 @@ defmodule Radiator.Directory.Editor.Manager do
   end
 
   def create_audio(network = %Network{}, attrs) do
-    Ecto.build_assoc(network, :audios)
-    |> Audio.changeset(attrs)
-    |> Repo.insert()
+    Multi.new()
```
monkeydom: wondering if we should change that to be `create_audio_publication` instead now. And one only can create an episode or an audio_publication and the audio is implied.

eteubert: I'm doing #178 next and will keep this in mind. 

We could `create_audio/2` in Podcast/Network with Episode/AudioPublication as side-product to keep it consistent.

Or get rid of `create_audio` entirely and let `create_episode`/`create_audio_publication` create the audio?

The API workflow in either case would be to create an audio immediately in one of both publication contexts.

PR #190, opened 2019-07-30
feature(api, content): contribution REST and GraphQL API

### LINKED ISSUE #139: People/Contributions: REST API

monkeydom:
documentation in `rest_api.md` and `graphql_api.md`

fixes #139 
needs ecto.reset

### DIFF
--- a/lib/radiator/contribution/podcast_contribution.ex
+++ b/lib/radiator/contribution/podcast_contribution.ex
@@ -1,6 +1,7 @@
 defmodule Radiator.Contribution.PodcastContribution do
   use Ecto.Schema
   import Ecto.Changeset
+  alias Radiator.Repo
 
   schema "podcast_contributions" do
     field :position, :float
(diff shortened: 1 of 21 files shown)


lib/radiator/directory/editor.ex
```diff
@@ -736,4 +748,88 @@ defmodule Radiator.Directory.Editor do
...
+          a.position < b.position
+
+        true ->
+          a.role_id < b.role_id
+      end
+    end)
+  end
+
+  def create_contribution(actor, subject, attrs) do
+    with :ok <- with_permission(actor, subject, :edit) do
```
eteubert: I like your introduction of `with_permission/3`, creates nice code flows 👍

lib/radiator/directory/editor.ex
```diff
@@ -736,4 +748,88 @@ defmodule Radiator.Directory.Editor do
...
+
+  def delete_contribution(actor, id) do
+    with {:ok, contribution, subject} <- get_contribution_and_subject(actor, id),
+         :ok <- with_permission(actor, subject, :edit) do
+      Repo.delete(contribution)
+    end
+  end
+
+  defp get_contribution_and_subject(actor, id) do
+    case Repo.get(AudioContribution, id) || Repo.get(PodcastContribution, id) do
```
eteubert: Does this not have an edge case where both ids are identical, making both conditions true, making it impossible to get the PodcastContribution because the AudioContribution is always returned first?

eteubert: Ah, saw the migration now, which ensures unique IDs over both tables.

lib/radiator/contribution/podcast_contribution.ex
```diff
@@ -15,6 +16,25 @@ defmodule Radiator.Contribution.PodcastContribution do
...
     contribution
-    |> cast(attrs, [:position])
+    |> cast(attrs, [:position, :person_id, :role_id, :podcast_id])
+    |> foreign_key_constraint(:person)
+    |> foreign_key_constraint(:role)
+    |> foreign_key_constraint(:podcast)
+    |> ensure_position()
+  end
+
+  def ensure_position(changeset) do
```
eteubert: So you're WET here, not DRY? ;)

lib/radiator/directory/editor.ex
monkeydom: Yeah - went through the same process ^^

lib/radiator/contribution/podcast_contribution.ex
monkeydom: yeah currently. As maybe this could be also just a default from the db in the migration? didn't have the time to choose the right one, so did it this way.

PR #191, opened 2019-07-31
Issue/178 audio and publish states

### LINKED ISSUE #178: REST API: Clean up audio & published states

eteubert:
I had some back and forth here on how to expose this via APIs as there are multiple ways and I'm not sure which is nicer. Here's how it's right now:

- Audios are created via the Audio REST endpoint with either an episode id or network id as parameter
- AudioPublications cannot be created, they are side products of the above

which is strange, because to make it behave the same way, creating an episode should be done by creating an audio via podcast id, not episode id? Or the other way: treat AudioPublications more like "Network Episodes"?

I also played with exposing publish date/state via GQL directly in the audio element but that felt confusing. Now I expose the data as it is, in Episode and AudioPublication.

In summary, it should fulfill the UI/Frontend requirements but I'm unsure about the general API design.

fixes #178


lib/radiator_web/router.ex
```diff
@@ -94,6 +94,9 @@ defmodule RadiatorWeb.Router do
 
     resources "/networks", NetworkController, only: [:show, :create, :update, :delete] do
       resources "/collaborators", CollaboratorController, only: [:show, :create, :update, :delete]
+
+      resources "/audio_publications", AudioPublicationController,
```
monkeydom: since episodes are not under `podcasts` and `podcasts` are not under `episodes` I would argue that audio publications shouldn't be either in the api

eteubert: Searched around a bit for arguments / best practices regarding nesting. There are various opinions out there, but there seems to be a general trend towards "you can't go wrong if you nest less". Here are some arguments why nesting can be annoying: https://stackoverflow.com/a/36410780/72448

An interesting approach from the same SO thread: https://stackoverflow.com/a/32820784/72448 Making objects accessible without nesting but nesting the CREATE for required relations. So for example you would access an episode through `GET /episodes/:id` but create it  via `POST /podcasts/:pid/episodes`. Surely not something to tackle right now but I find it an interesting approach.

Back to the project at hand: will pull out the route as you suggested.

monkeydom: Ack. I would like that. If the client apis don't suck I do think nesting the create only consistently seems good here. Also would remove the weird either one or the other id is required bits.

Issue #175, opened 2019-07-27
GQL Chapters API: include duration


eteubert:

eteubert:
included in #194 via https://github.com/podlove/radiator/pull/194/commits/271fd48f3e7fdf1c4a09ada136f680803cc192f8

PR #194, opened 2019-08-01
#167 REST API: Chapters Converter

### LINKED ISSUE #167: REST API: chapter import
Import various chapter files to an audio.

eteubert:
An endpoint that takes chapters data and converts it to the given format. No storage is touched, purely data conversion.

**Notes:**

1. Do we still actually need the endpoint that touches the database? In theory the Frontend could use this converter endpoint and then create each chapter individually via the existing chapter API.
2. I am "guessing" the chapter format by looking at the data format. I believe we should not expect the API consumer to provide the supplied format. My current guessing logic is quite naîve but can be made more robust as required.
3. I used an Ecto embedded schema in the controller for user input validation, feels like a good tool for the job.

closes #167


lib/radiator/audio_meta/chapter_import.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Radiator.AudioMeta.ChapterImport do
+  @spec import_chapters(binary()) :: {:ok, [Chapters.Chapter.t()]} | {:error, any()}
+  def import_chapters(data) do
+    format = guess_format(data)
```
monkeydom: I would say guess format should move to the chapters library eventually.

guides/rest_api.md
```diff
@@ -656,3 +657,49 @@ You should delete a task after it is done. Eventually done tasks will clear out
...
+Content Type: `application/json; charset=utf-8`
+
+The output is a list of chapters with the following keys:
+
+| Name    | Type      | Description                        |
+| ------- | --------- | ---------------------------------- |
+| `image` | `string`  | chapter image URL. null if empty   |
+| `link`  | `string`  | null if empty                      |
+| `start` | `integer` | chapter start time in milliseconds |
+| `title` | `string`  | chapter title                      |
```
monkeydom: Hmm. In my head this was different:
a) there was code point that would return chapters in one of the formats e.g. `GET /api/rest/v1/audios/:audio_id/chapters` would grow an optional format argument that would return json, psc, mp4chaps or the database format from the db
b) the convert code path would only ever return our database format sans the audio_id. Which incidentally this is, sans `start_string` and `duration`

Also I do think you probably should put in the `duration` and `duration_string` to the output as well if you are in the vicinity.


P.S.: 
There is also a docu bug in the rest_api file with chapters update (the last entry of chapter on the update section should be `chapter[image]` type `file`.

lib/radiator_web/controllers/api/preview/chapters_import_controller.ex
```diff
@@ -0,0 +1,61 @@
...
+
+    def changeset(params) do
+      %__MODULE__{}
+      |> cast(params, [:data, :output_format])
+      |> validate_required(:data)
+      |> validate_inclusion(:output_format, ["json", "psc", "mp4chaps"],
+        message: ~s(must be one of: json, psc, mp4chaps)
+      )
+    end
+  end
```
monkeydom: Yes, that embedded ecto schema for validation looks nice.

guides/rest_api.md
monkeydom: Referencing #175 here for easy closing.

eteubert:
Should be as you described now. Makes more sense this way.

PR #196, opened 2019-08-07
Issue/127 Fix Tracking

### LINKED ISSUE #127: Fix episode download tracking
Tracking now explicitly needs either an episode or network. At the moment only the audio file is given in the URL, which is not enough any more. While we're at it, we should make use of the new public URLs.

- maybe move TrackingController to `RadiatorWeb.Public`?
- route something like: `get "/:podcast_slug/:episode_slug/file/:file_id", TrackingController, :track`
- adjust controller to take the new parameters, see `Radiator.Tracking.track_download/1` for expected parameters

eteubert:
- updated tracking URLs and made them work again
- added AudioPublication to tracking
- add slug logic to AudioPublication
- update `AudioFile.public_url` (tracking URL) generation

closes #127

### DIFF
--- a/lib/radiator/directory/audio_publication.ex
+++ b/lib/radiator/directory/audio_publication.ex
@@ -6,12 +6,14 @@ defmodule Radiator.Directory.AudioPublication do
 
   alias Radiator.Directory.{
     Network,
-    Audio
+    Audio,
+    TitleSlug
   }
 
-  # todo: needs a slug, like episodes
   schema "audio_publications" do
     field :title, :string
+    field :slug, TitleSlug.Type
+
     field :publish_state, Radiator.Ecto.AtomType, default: :drafted
     field :published_at, :utc_datetime
 
@@ -28,10 +30,14 @@ defmodule Radiator.Directory.AudioPublication do
     audio_publication
     |> cast(attrs, [
       :publish_state,
-      :audio_id
+      :audio_id,
+      :title,
+      :slug
     ])
     |> validate_publish_state()
     |> maybe_set_published_at()
+    |> TitleSlug.maybe_generate_slug()
+    |> TitleSlug.unique_constraint()
   end
 
   @doc """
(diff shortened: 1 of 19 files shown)


lib/radiator/directory/audio_publication.ex
```diff
@@ -27,10 +30,14 @@ defmodule Radiator.Directory.AudioPublication do
...
       :publish_state,
-      :audio_id
+      :audio_id,
+      :title,
+      :slug
     ])
     |> validate_publish_state()
     |> maybe_set_published_at()
+    |> TitleSlug.maybe_generate_slug()
+    |> TitleSlug.unique_constraint()
```
monkeydom: Shouldn't that just be scoped to a network?

lib/radiator/media/audio_file.ex
```diff
@@ -44,12 +50,50 @@ defmodule Radiator.Media.AudioFile do
...
+    )
   end
 
-  def tracking_url(audio_file = %__MODULE__{}) do
-    RadiatorWeb.Router.Helpers.tracking_url(RadiatorWeb.Endpoint, :show, audio_file.id)
+  def tracking_url(audio_file = %__MODULE__{}, audio_publication = %AudioPublication{}) do
+    RadiatorWeb.Router.Helpers.tracking_url(
+      RadiatorWeb.Endpoint,
+      :track_audio_publication_file,
+      audio_publication.slug,
```
monkeydom: I'd like to see network in here to, so audio publications aren't scoped globally for one radiator instance.

lib/radiator/directory/audio_publication.ex
eteubert: You mean representing an AudioPublication as network+audio in URLs, so we never need the title slug? My thought was that since it's very similar to an episode, it should behave in a similar way.

monkeydom: No. Network slug and Audio slug.

eteubert: Audio doesn't have a title/slug 🤔

monkeydom: I meant network/slug and audiopublication/slug - sorry.

PR #198, opened 2019-08-07
feat(rest-api): when creating audio, allow params for episode/audio publication

eteubert:
Via discussion in Slack: make it easier for API consumer to create an audio with episode / audio publication by taking parameters for the episode / audio publication while creating the audio.

/cc @Fischaela

### DIFF
--- a/lib/radiator_web/controllers/api/audio_controller.ex
+++ b/lib/radiator_web/controllers/api/audio_controller.ex
@@ -3,10 +3,16 @@ defmodule RadiatorWeb.Api.AudioController do
 
   alias Radiator.Directory.Editor
 
-  def create(conn, %{"network_id" => network_id, "audio" => params}) do
+  def create(conn, params = %{"network_id" => network_id}) do
     with user = current_user(conn),
          {:ok, network} <- Editor.get_network(user, network_id),
-         {:ok, audio} <- Editor.create_audio(user, network, params) do
+         {:ok, audio} <-
+           Editor.create_audio(
+             user,
+             network,
+             Map.get(params, "audio"),
+             Map.get(params, "audio_publication", %{})
+           ) do
       conn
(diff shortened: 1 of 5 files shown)


lib/radiator_web/controllers/api/audio_controller.ex
```diff
@@ -3,21 +3,33 @@ defmodule RadiatorWeb.Api.AudioController do
 
   alias Radiator.Directory.Editor
 
-  def create(conn, %{"network_id" => network_id, "audio" => params}) do
+  def create(conn, params = %{"network_id" => network_id}) do
```
monkeydom: Either documentation is wrong, or you are matching wrong. (the "network_id" key is still inside an "audio")

lib/radiator_web/controllers/api/audio_controller.ex
```diff
@@ -3,21 +3,33 @@ defmodule RadiatorWeb.Api.AudioController do
...
+           ) do
       conn
       |> put_status(:created)
       |> put_resp_header("location", Routes.api_audio_path(conn, :show, audio))
       |> render("show.json", %{audio: audio})
     end
   end
 
-  def create(conn, %{"episode_id" => episode_id, "audio" => params}) do
+  def create(conn, params = %{"episode_id" => episode_id}) do
```
monkeydom: dito. Maybe a test added would be good?

lib/radiator_web/controllers/api/audio_controller.ex
eteubert: These (`network_id` and `episode_id`) are not params supplied but the client but the ones extracted from the route. Comes from the nesting under network/episode resource.

monkeydom: Ah, some magic.

Issue #200, opened 2019-08-18
Refactor User/Meta/People Architecture


eteubert:
- kill 1:1 between `Person` and `User`
- keep only authentication related things in `User`
- for other account or user management related matters, keep metadata in a `UserMeta` table
- add 1:n between `User` and `Person` because a user can be represented in different ways by different instances of a `Person` -- for example across different networks or simply to express different personas in different podcasts / episodes (Eric the coffee nerd, Eric the Elixir guy, Eric the dad, ...)

monkeydom:
Yeah. Maybe change `UserMeta` to `UserProfile`? but generally fits the bill for me.

PR #203, opened 2019-08-20
refactor architecture: User/Person/UserProfile

eteubert:
I detached People from Users. Instead, a UserProfile is now created together with each user.

UserProfile currently holds the `display_name` and `image`. `User` now only holds fields that are required for Auth stuff.

I wonder if we should expose this separation of User and Profile in the APIs at all? My gut says no, only expose a single "user" object in REST/gql as this separation is mainly an architecture choice to keep Auth.User clean.

I put UserProfile in `Radiator.Directory` because it is kind of public information? Otherwise we would probably a new context because neither does it fit into `Auth` nor `Contribution`.

⚠️ requires `mix ecto.reset`

issue #200

### DIFF
--- a/lib/radiator/auth/register.ex
+++ b/lib/radiator/auth/register.ex
@@ -56,24 +56,14 @@ defmodule Radiator.Auth.Register do
     |> Repo.all()
   end
 
-  # TODO: we want to always have a person associated with a user but can we do it without coupling the logic here?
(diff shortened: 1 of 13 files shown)


monkeydom:
My gut agrees with the assesment to not expose this to the api (for now). We might expose this at a later time when we add additional authentication mechanisms for the same user.

PR #206, opened 2019-08-25
Tracking Generation, Download Cleaning, Job Management

eteubert:
### Main Features

Adds `RadiatorSupport.TrackingGenerator` to generate/simulate download data. In previous projects I repeatedly thought “I should have a download generator” because without it, building reporting tools and finding performance bottlenecks is hard. So finally, this time, I built it before starting anything else.

Adds `Radiator.Tracking.Cleaner` which removes duplicate requests from the database.

Removes `clean` flag from tracking data because I figured we don’t need it. Instead of marking an entry as “not clean”, we can as well delete it.

### Behind the Scenes

Adds [Oban](https://github.com/sorentwo/oban) as job processor. I’ve been trying it out at BR and am very happy with it so far. Currently only used in `Radiator.Tracking.CleanerWorker`.

Adds [Quantum](https://github.com/quantum-elixir/quantum-core) as cron-like scheduler. I’ve been using it for a long time at BR and it’s doing its job very well. Includes stuff like cluster-awareness. Currently used to schedule tracking cleaning on a daily basis.


lib/radiator_support/generator/user_agents_downloader.ex
```diff
@@ -0,0 +1,25 @@
+defmodule RadiatorSupport.UserAgentsDownloader do
+  @moduledoc """
+  Downloads file with 10.000 user agents for RadiatorSupport.TrackingGenerator
+  because keeping a 1MB file with noncritical data versioned here feels wrong.
```
monkeydom: nah - just put it in. git zips resources anyways. And having this weird extra dependenchy on a gist isn't better imho.

PR #211, opened 2019-09-08
feat(deployment): local deployment

monkeydom:
* switch to use alpine instead of slim for smallest docker images
* preparing for use with [radiator-deploy](https://github.com/podlove/radiator-deploy)
* preparing for actual manual deployment behind nginx out in the world with https (in progress)
* move docker-compose into a subdirectory to be able to use `.env` file without also polluting the environment of the regular `Dockerfile`


monkeydom:
@eteubert I guess this change also needs an adjustment for the drone.io docker build. Please advise how to adjust properly.

Also this is work in progress, but spent enough time on it to want to pull it in. Next steps are:
- add the actuall cms to the `radloc` compose
- add a first use screen
- prepare another compose file for manual deployment behind an existing https nginx infrastructure

PR #240, opened 2019-09-25
cleanup publication state API

### LINKED ISSUE #237: Podcast Publish State
- add podcast.publishState to GQL (only isPublished)
- remove podcast.isPublished from GQL
- document publishState for podcast in API doc

eteubert:
- gql: remove isPublished from episode and podcast
- gql: both episode and podcast now have a publishState accessor returning "drafted", "scheduled", "published" or "unpublished"
- REST: both episode and podcast now support the "publish_state" field on UPDATE/CREATE

fixes #237

### DIFF
--- a/guides/rest_api.md
+++ b/guides/rest_api.md
@@ -241,6 +241,7 @@ DELETE /api/rest/v1/networks/:id/collaborators/:username
 | `podcast[owner_name]`  | `string`  |                                                                                       |
 | `podcast[owner_email]` | `string`  |                                                                                       |
 | `podcast[slug]`        | `string`  |                                                                                       |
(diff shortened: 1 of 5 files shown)


eteubert:
@monkeydom With the gql api we had dedicated publish/depublish mutations but I wonder if this is necessary? Isn't this an attribute like any other? Would you prefer dedicated REST routes to publish/depublish/schedule podcasts&episodes which use the Editor.publish/depublish Elixir API instead of setting the state directly?

monkeydom:
I kind of dislike for the clients to know the special meaning of changing the state/date pair here. especially for scheduled. however, newer api paths don't look nice either, do they?

PR #242, opened 2019-09-26
Issue/233 sandbox mode

### LINKED ISSUE #233: [User Story] Global DEMO flag/mode
**As an** administrator
**I want** to set a radiator instance into DEMO mode

#### Acceptance Criteria

1. all feeds contain `<itunes:block>Yes</itunes:block>` (see https://help.apple.com/itc/podcasts_connect/?lang=en#/itcb54353390)
1. DEMO flag can be accessed via API
1. generate robots.txt forbidding all indexing

eteubert:
A global sandbox mode for the Radiator instance.

  The following changes apply to a system in sandbox mode:

  - all feeds contain `<itunes:block>Yes</itunes:block>`
  - a robots.txt is generated disallowing all requests

  Set via config

  ```
  config :radiator, :sandbox_mode, enabled: false
  ```

  Or in releases per ENV by setting `SANDBOX_MODE_ENABLED` to true.

  Status can be queried by clients:

  ```
  query {
    sandboxMode {
      enabled
    }
  }
  ```

closes #233


config/dev.exs
```diff
@@ -91,3 +91,5 @@ config :radiator, Radiator.Repo,
   pool_size: 10
 
 config :radiator, Radiator.Mailer, adapter: Radiator.Email.Console
+
+# config :radiator, :sandbox_mode, enabled: true
```
monkeydom: ist auch gut für dev aktiv eigentlich, oder?

lib/radiator/sandbox_mode.ex
```diff
@@ -0,0 +1,31 @@
...
+    sandboxMode {
+      enabled
+    }
+  }
+  ```
+  """
+  def enabled? do
+    Application.get_env(:radiator, :sandbox_mode)[:enabled] == true
+  end
+end
```
monkeydom: I this the way to go? Aka. generate a top level module to access the config? Or Should we have a more general `Radiator.AppConfig` where we collect these things? (Asking for the redirect to the cms, which I would have done by introducing another config var.

eteubert: Felt good in this case because this way I have a place to put all the documentation for the feature switch. Initially I planned to add sub-switches (block on/off, robots on/off etc.) but that seemed unnecessary later.

I don't mind having a separate module for each feature switch, possibly in a subdirectory to not clutter the root directory.

Yours doesn't sound like a feature switch though, just a config. Not sure if all config access should be wrapped in a separate module. But may be worth a try.

monkeydom:
I had to fix this in a later commit, all environment variables are always strings so.

PR #253, opened 2019-09-27
fix: automatically extract duration and file format from uploaded audio file

### LINKED ISSUE #176: Audio File Upload: Automatically extract duration

monkeydom:
* added duration and `audio_format` to `audio_file`
* enforced all audio file creation to walk the same path
* automatically set the duration of an audio on upload
* update rest api now only allows for title changes, to exchange the actual audio file one has to delete it and create a new one
* exposed the new properties in graphql

fixes #176


lib/radiator/media/audio_file_upload.ex
```diff
@@ -58,30 +64,74 @@ defmodule Radiator.Media.AudioFileUpload do
...
-      })
+  def probe_file(path) do
+    with {:ok, streams} <- FFprobe.streams(path) do
+      stream =
+        streams
+        |> Enum.find(hd(streams), fn m -> m["codec_type"] == "audio" end)
+
+      duration = ((Float.parse(stream["duration"]) |> elem(0)) * 1_000) |> ceil
+
+      audio_format =
```
eteubert: How about a file_meta column of type map where we can store arbitrary key-value-pairs? Easier to work with than comma separated values as API/frontend might want to display these metas in their own way.

monkeydom: yeah - a more clearly defined storage of the extracted metadata is probably prudent. This is more intended as a read only value for now and as a finger excersize in terms of what FFprobe is delivering.

PR #267, opened 2019-10-09
feature: feeds as files

### LINKED ISSUE #259: Feeds as file artefacts
At the moment feeds are generated on-the-fly. Instead, they should be stored and delivered as files.

- [x] generate feed and store as xml file on disk
- [x] keep dynamic feed URL (so we can do feed request tracking) but deliver static file
- [x] support paged feeds (multiple files per feed)
- [x] use jobs to generate feed files
- [x] regenerate feed file every time something in podcast or any episode changes
- [ ] what to do with feed requests when there is no feed file yet?
- [ ] cleanup feed files (last pag(es) might be zombie files if for some reason the actual number of pages of a podcast decreases)

eteubert:
Instead of serving feeds on-demand, static feed files are generated every time podcast or episode data changes, and stored on disk. Public feed URLs are dynamic and incoming requests redirected to the static file, similar to how audio files are handled.

closes #259

### DIFF
--- a/lib/radiator/directory/editor/manager.ex
+++ b/lib/radiator/directory/editor/manager.ex
@@ -16,7 +16,8 @@ defmodule Radiator.Directory.Editor.Manager do
     Podcast,
     Episode,
     Audio,
-    AudioPublication
+    AudioPublication,
+    Publication
   }
 
   alias Radiator.Contribution.{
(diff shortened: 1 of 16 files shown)


lib/radiator/directory/editor/manager.ex
```diff
@@ -269,9 +300,14 @@ defmodule Radiator.Directory.Editor.Manager do
...
   def publish(subject = %type{}) when type in [Podcast, Episode, AudioPublication] do
-    subject
-    |> type.changeset(%{publish_state: :published})
-    |> Repo.update()
+    result =
+      subject
+      |> type.changeset(%{publish_state: :published})
+      |> Repo.update()
+
+    refresh_feed_cache(subject)
```
monkeydom: This recurring call probably wants some clearer language concept around it, so it is not forgotten when it is needed.

eteubert: I'm not happy with it as well. That pattern where I have to keep the result in a tmp variable so I can still use it as the function return feels very awkward.

```
tmp = do_something()
refresh_feed_cache(subject)
tmp
```

Not sure what a clean solution would look like.

Issue #497, opened 2023-11-17
Set basic data structure (station, podcast, episode)


sorax:

sorax:
Hey @lukad & @electronicbites,

could you please have a look at the idea for structuring data?
Maybe we could have a quick discussion about this?

electronicbites:
Since we have identified a very variable definitions of items I would like to stay as flexible as possible at least in the beginning until we gathered some experience. We are probably dealing with a tree structure building the outliner. So every leave in the tree has a type, a parent node, a public uuid and a type dependend list of attributes. We could start with a flexible Json structure but we might end with an extra table for these, e.g. `URLItem` or likely. 
But the tree consists of generic items, so we can add URLs to an episode or a chapter (correct name?) of an episode.

electronicbites:
For podcasts I would suggest a state-machine. It has different states and transitions (e.g. `publish`).

lukad:
@electronicbites, I think having a seperate table for each item type is the way to go in terms of maintainability and db operations. An `item` would have a `type`/`kind` set to `url` + `url_item_id`.

State machine per might be a good idea, I can think of at least one other state which would be something like `on-air`, although that would be on an episode basis.

I don't really understand the reason for always having a "wip" episode. Why not just create it when it's needed?

PR #501, opened 2023-11-28
Add base resources

electronicbites:
Add basic resources stations, podcasts and episodes 
related to [#497](https://github.com/podlove/radiator/issues/497)

### DIFF
--- a/lib/radiator/podcast.ex
+++ b/lib/radiator/podcast.ex
@@ -1,0 +1,293 @@
+defmodule Radiator.Podcast do
+  @moduledoc """
+  The Podcasts context.
+  Handles repo operations for networks, shows and episodes.
+  """
+
+  import Ecto.Query, warn: false
+  alias Radiator.Repo
+
+  alias Radiator.Podcast.{Episode, Network, Show}
+
+  @doc """
+  Returns the list of networks.
+
+  ## Examples
+
+      iex> list_networks()
+      [%Network{}, ...]
+
+  """
+  def list_networks do
+    Repo.all(Network)
+  end
+
+  @doc """
+  Gets a single network.
+
+  Raises `Ecto.NoResultsError` if the Network does not exist.
+
+  ## Examples
+
+      iex> get_network!(123)
+      %Network{}
+
+      iex> get_network!(456)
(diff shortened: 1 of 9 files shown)


lib/radiator/podcasts.ex
```diff
@@ -0,0 +1,297 @@
+defmodule Radiator.Podcasts do
```
bhtabor, line 1: Could Broadcast be a better context name than Podcasts? [Contexts are dedicated modules that expose and group related functionality.](https://hexdocs.pm/phoenix/contexts.html#thinking-about-design)

electronicbites: Thanks for the input. I agree, I am not satisfied with the naming of the context `Podcasts` as well. It should differentiate to the new outliner tool, we are working on, the media files a podcast will be associated with. It is the base structure for all phases of the app. Our core. For my feeling `Broadcasts` might be a bit misleading because especially in phase 1 the podcast has not been broadcasted yet. 
Naming is so hard ;)

bhtabor: Yeah. :) `Podcasts` is fine as a placeholder then.

sorax: How about the name **Show**? (Tim suggested that once)
So the hierarchy would be:
1. Podcast.Network (e.g. Metaebene or ARD or BBC)
2. Podcast.Show (e.g. Freakshow or Minkorrekt)
3. Podcast.Episode (e.g. FR123 or Mi014)

sorax:
Hey @electronicbites , would you delete the `views` & the `page_html`?
I think we don't need them and can merge this branch afterwards.

sorax (changes requested):
question: can we delete the views & the page_html?

electronicbites:
> question: can we delete the views & the page_html?

of course, but it was not part of my plan to include the change in this PR. I oversaw the movement. Fixed that. Since this is unrelated I will make an extra PR for this change ;)

electronicbites:
@sorax now the PR is ready for review, thanks ;)

sorax (approved):
Everything fine! Feel free to merge.

PR #502, opened 2023-11-28
Basic seeds

electronicbites:
very basic seeds
but they also show some current missing pieces, like outlines need to be connected to episodes, there needs to be a current episode and more .. 
 
based on https://github.com/podlove/radiator/pull/501


sorax (approved):
looks good to me

PR #506, opened 2023-12-06
add credo

electronicbites:
I would like to add credo to  maintain a common code style guideline ... any objections?


sorax:
Totally agree!
I would really appreciate having credo.

PR #510, opened 2023-12-16
Connect outline nodes to episodes

electronicbites:
Connect nodes to episodes. This is a required argument. No node should be be dangling since they would never be shown. 

The controller expects a show_id for a node to be created. In order to connect the node the a episode we need do find a episode. In the near future this will be the one in draft mode, or even the one stored in a user preference. 
Currently it just returns the episode with the highest number. And episodes have a number now.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019 Podlove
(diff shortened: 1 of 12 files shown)


sorax (approved):
looks good to me

PR #511, opened 2023-12-19
remove not needed view files

electronicbites:
@sorax seems like removing the views wont work . they are referenced ....


sorax:
the "view" files aren't in the current main branch. so everything fine so far, I would say.
can we close this?

electronicbites:
yes of course. perhaps this was also a misunderstanding. I tried to move the existing view files into its own folder. But then we discussed wether we need them at all. So currently there are these view files in the controllers folder, like error and page html. 
anyway this was just an experiment becaue we discussed that, I will close this PR to not confuse as any further ;)

PR #548, opened 2024-07-30
Update to elixir 1 17 2 & update dependencies

electronicbites:
again upgrade elixir, but now with a compatible (updated) debian version


sorax (approved):
go for it

PR #558, opened 2024-09-01
Feature/url extractor

electronicbites:
Introduces a mechanism to detect URLs in node contents. If an URL is detected it just gets logged. It should be stored as an entity to the database. The frontend could present detected URls, the content of a website could be scraped and URLs can be exported to the show notes

This PR also introduces ChangesListener that subscribes to every change in the outliner.
The `EventConsumer` was renamed to  `CommandProcessor`

episode_id was added to every event type while it was removed from the `Dispatch.subscribe` function since it was not used.

### DIFF
--- a/lib/radiator/application.ex
+++ b/lib/radiator/application.ex
@@ -7,6 +7,7 @@ defmodule Radiator.Application do
 
   alias Radiator.Outline.CommandProcessor
   alias Radiator.Outline.EventProducer
+  alias Radiator.Outline.NodeChangeListener
 
   @impl true
   def start(_type, _args) do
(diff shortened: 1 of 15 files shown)


sorax:
Hey @electronicbites, regarding the topic:
> "episode_id was [...] removed from the Dispatch.subscribe function since it was not used."

when we have multiple outlines, how can we only subscribe (get updates for) the outline we are currently editing and not updates from other episode-nodes someone else is using in another episode?

is this already the case or would the subscribe with episode_id be a good idea here?

sorax (approved):

sorax:
let's merge this and do it later on

PR #561, opened 2024-09-05
Optional parent node for move node

electronicbites:
I had to change the signature for `move_node` otherwise it would not be possible to differentiate between setting a value to `nil`and omitting the value. I used a keyword list since this a classic for optional parameters. Another option would be to use a map like for `insert_node`


sorax (approved):
looks good to me

PR #562, opened 2024-09-09
Indent  nodes

electronicbites:
Implement the indent node usecase as a command
result will not be preprocessed in the frontend, therefor we dont filter move node events. 
at least at my local machine this works fast enough.


lib/radiator_web/live/outline_component.ex
```diff
@@ -228,11 +228,9 @@ defmodule RadiatorWeb.OutlineComponent do
   defp indent(socket, uuid, prev_id) do
     user_id = socket.assigns.user_id
 
-    Dispatch.move_node(uuid, user_id, generate_event_id(socket.id),
-      parent_id: prev_id,
-      prev_id: nil
-    )
+    Dispatch.indent_node(uuid, user_id, generate_event_id(socket.id))
 
+    # TODO prev_id might not be nil!!
```
electronicbites, line 233: this is tricky. prev id might be the last element in the list (not nil) if new parent already contains children

PR #564, opened 2024-09-11
feature: display all editing users in different colors

mwingert:
- each user has it's own <span> tag
- each user name is displayed in one of 12 colors * 3 intensities = 36 color variations
- color selection is based on first char and user name length


sorax (approved):
looks good and today I learned about tailwind safelist

PR #571, opened 2024-09-21
implement `move_up`, `move_down`

electronicbites:
Implements simplified move commands for moving a node up and down (per keyboard command)


sorax (approved):
great improvement - thank you!

PR #572, opened 2024-09-22
fix node events

sorax:
@electronicbites could you please have a look at lib/radiator/outline/node.ex ?
from my point of view a uuid should only be set, of none is present (this wasn't the case)

the uuid was overwritten if the changeset was valid and a uuid was already preset (was set, but not changed)


lib/radiator/outline/node.ex
```diff
@@ -53,13 +53,10 @@ defmodule Radiator.Outline.Node do
...
 
-  defp put_uuid(%Ecto.Changeset{valid?: true} = changeset) do
-    if changed?(changeset, :uuid) do
-      changeset
-    else
-      put_change(changeset, :uuid, Ecto.UUID.generate())
+  defp put_uuid(%Ecto.Changeset{} = changeset) do
+    case get_field(changeset, :uuid) do
+      nil -> put_change(changeset, :uuid, Ecto.UUID.generate())
+      _ -> changeset
```
electronicbites, line 59: I see. But I do not understand. There are several tests in `node.exs` that do exactly test this behaviour. Could you provide a scenario which works now and did not work before? I just want to understand. But of course if thats the case and my tests are still green go for it ...,

electronicbites (commented):
added a question

lib/radiator/outline/node.ex
electronicbites: Look at this test: https://github.com/podlove/radiator/blob/d478ddb395561abf5ef886c90bde2f5d939d9f67/test/radiator/outline/node_test.exs#L11

sorax: additional test was needed to show the difference

PR #573, opened 2024-09-30
feature: Reload episodes after creating a new or updating an existing…

mwingert:
… episode


sorax (approved):
great improvement - makes perfect sense

PR #579, opened 2024-10-07
feature: Soft deleting episode

mwingert:
Marking episodes as `is_deleted = true` and setting `deleted_at = now`. 
`list_available_episodes` gives all episodes that are NOT marked as deleted. 
`list_all_episodes` gives all episodes including the once that are marked as deleted.


electronicbites (approved):
nice, looks good to me

electronicbites (approved):
tested it locally and it worked, therefor I will merge

PR #582, opened 2024-10-10
feature: editing existing shows

mwingert:
Each show has an edit button that displays the show form and allows to make changes on the title, description and the hosts:

Edit button of show

Obviously the styling could be improved, I would leave that to a future PR.

> [!IMPORTANT]
review: Please give the function `defp update_associate_hosts(show, hosts) do`  a proper look, if you have a better idea on how to handle the deletion of `ShowHosts`, let me know.

### DIFF
--- a/lib/radiator/podcast.ex
+++ b/lib/radiator/podcast.ex
@@ -7,7 +7,7 @@ defmodule Radiator.Podcast do
   import Ecto.Query, warn: false
   alias Radiator.Repo
 
-  alias Radiator.Podcast.{Episode, Network, Show}
+  alias Radiator.Podcast.{Episode, Network, Show, ShowHosts}
 
   @doc """
   Returns the list of networks.
@@ -185,6 +185,39 @@ defmodule Radiator.Podcast do
     %Show{}
     |> Show.changeset(attrs)
     |> Repo.insert()
+  end
+
+  @doc """
+  Creates a show with hosts.
+
+  ## Examples
+
+      iex> create_show(%{field: value}, [%User{}])
+      {:ok, %Show{}}
+
+      iex> create_show(%{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def create_show(attrs, hosts) do
+    Repo.transaction(fn ->
+      case create_show(attrs) do
+        {:ok, show} ->
+          associate_hosts(show, hosts)
+          show
+
+        {:error, changeset} ->
+          Repo.rollback(changeset)
+      end
+    end)
+  end
+
+  defp associate_hosts(show, hosts) do
+    Enum.each(hosts, fn host ->
+      %ShowHosts{}
+      |> ShowHosts.changeset(%{show_id: show.id, user_id: host.id})
+      |> Repo.insert!()
+    end)
   end
 
   @doc """
--- a/lib/radiator/podcast.ex
+++ b/lib/radiator/podcast.ex
@@ -216,9 +216,7 @@ defmodule Radiator.Podcast do
   # Suggested to run inside a transaction.
   defp associate_hosts(show, hosts) do
     Enum.each(hosts, fn host ->
-      %ShowHosts{}
-      |> ShowHosts.changeset(%{show_id: show.id, user_id: host.id})
-      |> Repo.insert!()
+      create_showhosts!(%{show_id: show.id, user_id: host.id})
     end)
   end
 
@@ -271,9 +269,7 @@ defmodule Radiator.Podcast do
 
     # Add users to hosts
     Enum.each(add_hosts, fn host ->
-      %ShowHosts{}
-      |> ShowHosts.changeset(%{show_id: show.id, user_id: host.id})
-      |> Repo.insert!()
+      create_showhosts!(%{show_id: show.id, user_id: host.id})
     end)
 
     # Remove users from hosts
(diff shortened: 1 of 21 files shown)


sorax:
Great work!

Hey @electronicbites could you please have a look at `update_associate_hosts/2` ?
I think you have a better understanding on this.

mwingert:
### Here is an alternative solution: 
By fetching the ShowHosts first and the delete them.
The code looks a bit nice but it fetches the data which is kinda unnecessary.

```elixir
  defp update_associate_hosts(show, hosts) do
    show = Repo.preload(show, :hosts)

    # intersection contains unchanged hosts -> ignore them
    {remove_hosts, add_hosts} = remove_intersections(show.hosts, hosts)

    # Add users to hosts
    Enum.each(add_hosts, fn host ->
      create_showhosts!(%{show_id: show.id, user_id: host.id})
    end)

    # Remove users from hosts
    remove_host_ids = Enum.map(remove_hosts, fn host -> host.id end)

    if !Enum.empty?(remove_host_ids) do
      from(s in ShowHosts,
        where:
          s.show_id == ^show.id and
            s.user_id in ^remove_host_ids
      )
      |> Repo.all()
      |> Enum.each(fn showhost -> delete_showhosts!(showhost) end)
    end
  end

  def delete_showhosts!(%ShowHosts{} = showhost) do
    Repo.delete!(showhost)
  end
```

If @electronicbites thinks it's more favorable then I will make the changes. It's already prepare for that case. :-)

electronicbites (commented):
I dont get it @mwingert, what is the difference between the alternate version and the one in the PR? 
But if you have a n:n association, you have two lists to maintain, one to delete and one to add. And thats exactly what you are doing. 
I just have the feeling it could be more compact. I will think over it ...

electronicbites:
In Ruby you could do something like 
```ruby
bad_hosts = existing_hosts - new_hosts
missing_hosts = new_hosts - existing_hosts
```

;)

electronicbites:
this also works for Elixir
```elixir
list1 = [1, 2, 3, 4, 5]
list2 = [3, 4, 6]

difference = list1 -- list2
```

lib/radiator/podcast.ex
```diff
@@ -212,14 +212,90 @@ defmodule Radiator.Podcast do
...
+    {list_a_without_intersection, list_b_without_intersection}
+  end
+
+  # Updates the list of associate hosts of a show by adding or removing users to hosts list.
+  # Suggested to run inside a transaction.
+  defp update_associate_hosts(show, hosts) do
+    show = Repo.preload(show, :hosts)
+
+    # intersection contains unchanged hosts -> ignore them
+    {remove_hosts, add_hosts} = remove_intersections(show.hosts, hosts)
```
electronicbites, line 268:
```elixir
remove_hosts = show.hosts -- hosts
add_hosts = hosts -- show.hosts
```

electronicbites: so you wont need the function `remove_intersections`

lib/radiator_web/live/admin_live/index.ex
```diff
@@ -125,6 +139,20 @@ defmodule RadiatorWeb.AdminLive.Index do
...
+    save_show(socket, socket.assigns.action, params)
+  end
+
+  def handle_event("delete", %{"id" => id}, socket) do
+    network = Podcast.get_network!(id)
+    {:ok, _} = Podcast.delete_network(network)
+
+    socket
+    |> assign(:networks, Podcast.list_networks(preload: :shows))
+    # |> stream_delete(:networks, network)}
```
electronicbites, line 151: why keep this commented line?

electronicbites (changes requested):
consider my suggestion to simplify the hosts calculation

lib/radiator_web/live/admin_live/index.ex
mwingert: Weird. I haven't changed the `handle_event("delete"`.

lib/radiator/podcast.ex
mwingert: Great point. Will make the change. Don't know why I didn't try that.

mwingert:
@electronicbites 
> I dont get it @mwingert, what is the difference between the alternate version and the one in the PR? But if you have a n:n association, you have two lists to maintain, one to delete and one to add. And thats exactly what you are doing. I just have the feeling it could be more compact. I will think over it ...

I meant the part: 
a)
```elixir
      count_removable_hosts = length(remove_host_ids)

      from(s in ShowHosts,
        where:
          s.show_id == ^show.id and
            s.user_id in ^remove_host_ids
      )
      |> Repo.delete_all()
      |> case do
        {^count_removable_hosts, nil} ->
          :ok

        {count_removed, _} ->
          Repo.rollback(
            "Couldn't remove all hosts (expect: #{count_removable_hosts} vs. #{count_removed})."
          )
```
vs:
b)
```elixir
      from(s in ShowHosts,
        where:
          s.show_id == ^show.id and
            s.user_id in ^remove_host_ids
      )
      |> Repo.all()
      |> Enum.each(fn showhost -> delete_showhosts!(showhost) end)
```

The former one (a) performs the deletion directly in the database, by creating a delete query, without fetching the data to Ecto first.
The latter one (b) fetches the ShowHosts first and then deletes each one.

electronicbites (approved):

PR #588, opened 2024-11-05
 add userservices model

electronicbites:
Introduces a new schema `web_services` which is supposed to store data for all kinds of possible external services. First implementation is for `Raindrop.io`. Hence the branchname the embedded data for the service it is not yet polymorphic. 
In the Admin Show settings a show can now be associated with a raindrop collection. 
Still missing: URL synchonization with an inbox folder.


sorax (approved):
looks fine

PR #590, opened 2024-11-12
switch events to contenteditable

sorax:


assets/js/hooks/events/handler.ts
```diff
@@ -2,39 +2,40 @@ import { NodeData, UserAction } from "../types";
...
+  setContent,
+  focusNode,
   addEditingUserLabel,
   removeEditingUserLabel,
 } from "../node";
 
 export function handleBlur({ uuid, user_name }: UserAction) {
-  const node = getNodeById(uuid)!;
-  removeEditingUserLabel(node, user_name);
+  // const node = getNodeById(uuid)!;
```
electronicbites, line 13: why not delete?

lib/radiator_web/live/outline_component.ex
```diff
@@ -154,128 +144,107 @@ defmodule RadiatorWeb.OutlineComponent do
...
+        "#{prev_node.content}#{content}",
+        user_id,
+        generate_event_id(socket.id)
+      )
+
+      Dispatch.delete_node(uuid, user_id, generate_event_id(socket.id))
+    end
 
     socket
+    # |> stream_delete_by_dom_id(:nodes, "nodes-form-#{uuid}")
```
electronicbites, line 191: can be deleted as well

lib/radiator_web/live/outline_component.ex
```diff
@@ -154,128 +144,107 @@ defmodule RadiatorWeb.OutlineComponent do
...
+        generate_event_id(socket.id)
+      )
+
+      Dispatch.delete_node(next_node.uuid, user_id, generate_event_id(socket.id))
+    end
 
     socket
-    |> stream_delete_by_dom_id(:nodes, "nodes-form-#{uuid}")
+    # |> stream_delete_by_dom_id(:nodes, "nodes-form-#{uuid}")
     |> reply(:noreply)
```
electronicbites, line 213: delete ?!

electronicbites (approved):
lg2m

except very small comments about code which could be deleted

assets/js/hooks/events/handler.ts
sorax: in the line below (line 14) we call: `removeEditingUserLabel(user_name);`
this function does remove the user-label (if exists):
```
removeEditingUserLabel(user_name: string) {
  const span = document.getElementById(user_name);
  span && span.remove();
}
```

lib/radiator_web/live/outline_component.ex
sorax: ✅ absolutely correct. code is removed now.

lib/radiator_web/live/outline_component.ex
sorax: ✅ absolutely correct. code is removed now.

PR #603, opened 2024-12-28
dispatch merge_prev & merge_next

sorax:


electronicbites:
@sorax can we merge this?

